PR #277, opened 2019-01-24
Added files for using docker and updated readme.

bobonthenet:


.docker/web/run.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
James-Firth: May as well exit early if any step fails, right? :) 
```suggestion
#!/bin/sh -e
```

bobonthenet:
Thanks for the suggestion. I'm a total noob with docker but I need to learn it for future projects at work. Any input is welcome.

James-Firth:
No problem, I use a lot of docker and bash stuff at work 😄 
You could also replace `step` with `echo` (auto-appends a `\n`) if you wanted.

I'm not familiar enough with apache or php to really comment on those parts, but they look fine to me!

pspeter3:
Was the intention for this container to be used in production or development? Also, I think some configuration/documentation is missing because it defaults to 404.

PR #280, opened 2019-01-28
Fix broken route in Attribute Templates calling a Location route when…

Enkrod:
… it should call the same Attribute Template


Enkrod:
This fixes a 404 in the Link "Entry" of Attribute Templates that show the Menu

PR #282, opened 2019-01-30
MinorFixes

Enkrod:
Slight UI improvements and a typo in DocBlock fixed


ilestis:
Assets are generated by webpack, so the resources/assets/sass files need to be edited, not the public/ ones.

PR #296, opened 2019-11-04
Proposal: bulk permissions editing

paulyhedral:
This pull request is a proposal to change some behavior of the bulk permissions editing dialog.

* The "override" boolean switch is removed from the dialog;
* Permission drop-downs include a third option, 'ignore', which replaces the "override" behavior. This new option becomes the default for each drop-down, so that a permissions change must be explicitly chosen.

Rationale: bulk-editing of permissions are usually performed in order to change a limited number of values, making a default of not changing a value easier for the admin to handle. When the list of subjects grows, having to select the desired value (and thereby remember what the original value is for a value that should not change) for each subject/permission combination becomes tedious and potentially error-prone, resulting in some undesired permissions values.

No code changes are included, since adding the 'ignore' action results in the (if-elseif) code path for altering permissions to skip over the 'add' and 'remove' behavior, as intended.

### DIFF
--- a/resources/lang/de/crud.php
+++ b/resources/lang/de/crud.php
@@ -251,6 +251,7 @@ return [
         'action'    => 'Aktion',
         'actions'   => [
             'bulk'          => [
+                'ignore'    => 'Ignorieren',
                 'add'       => 'Hinzufügen',
                 'remove'    => 'Entfernen',
             ],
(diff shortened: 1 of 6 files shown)


ilestis:
This makes so much more sense! Thanks :-)

PR #330, opened 2020-09-04
Typo: entitity

Salvatos:
I hope you don’t mind me PR’ing individual typos like this when I come across them; I feel like this would be more convenient for you than reporting them on Discord.


ilestis (approved):
Absolutely, thanks!

PR #333, opened 2020-09-07
Fixed broken link and grammar

Salvatos:
Also "All improvements, features and bugs must be related to a ticket on github" but tickets can’t be created?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ After cloning the project, create the following files.
 ## Using Vagrant
 Make sure you have [composer](https://getcomposer.org), [Vagrant](https://www.vagrantup.com/) and their dependencies installed and walk through the next steps:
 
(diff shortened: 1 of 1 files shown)


Salvatos:
Oh also is `DB_HOST=mysql`, which one is supposed to remove when using Docker, meant to refer to `DB_CONNECTION=mysql`?

ilestis (approved):
Thanks :) Not sure about the docker stuff as I don't use docker, sorry. Regarding tickets on github, that's an old reference as we do everything on trello now.

Salvatos:
Sorry, I meant that the parameters in the .env file are:
```
DB_CONNECTION=mysql
DB_HOST=127.0.0.1
...
```
But the instructions say to remove ```DB_HOST=mysql``` which is a mix of both.

And the reference to tickets should probably be removed in that case (or changed to whatever Trello uses).

PR #334, opened 2020-09-07
Show Organisations in Character side panel

Salvatos:
[Related Trello](https://trello.com/c/eWspOVZf/325-pin-organisation-membership)

I made a thing :3
You may not want it; that’s fine, I needed something simple to practice with.

You might want to:
- place it somewhere else in the sidebar,
- make a dedicated `characters.fields.organisations` string for it (I used the one from the tabs menu),
- rename some variables,
- make it optional or include roles as in ArcOnyx’s request?

I basically copied the new Notes sidebar list mixed with other bits.

Tested with 0, 1 and 3 organisations, as well as the Organisations module turned off; no visible breakage.

### DIFF
--- a/resources/views/characters/_menu.blade.php
+++ b/resources/views/characters/_menu.blade.php
@@ -18,6 +18,17 @@
                     <br class="clear" />
                 </li>
             @endif
+            @if ($campaign->enabled('organisations') && !$model->organisations->isEmpty())
+                <li class="list-group-item">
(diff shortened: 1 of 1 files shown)


ilestis:
I sadly have to reject this because our designer is working on a new way to show this kind of information altogether, and I'm trying to keep db queries low on the sidebar as there are already quite a lot of them. I also think you're not checking if the user has access (typically for a non-admin viewer of a campaign), which will end up in php errors because the organisation might be half empty. But nice effort!

PR #335, opened 2020-09-08
French translation fixes + rebase

graham2071:
I've reviewed all files in resource/lang/fr. Certainly missed some minor errors.


ilestis (approved):
Awesome, thanks! For future reference, the `voyager.php` file is a third party plugin so changes in the translation file will be erased by updating the plugin :)

PR #336, opened 2020-09-09
LGFM -> LFGM

Salvatos:
Typo in a few language files.


ilestis:
Fixing typos in EN/EN-US and FR are fine, but for other languages those get squashed every update by what's in the prod db so better reported directly to the translators.

PR #337, opened 2020-09-16
Bugfix/docker

graham2071:
The aim of this change to provide a build of kanka docker image and compose file to run kanka without the need to install anything (except docker). The docker build was outdated.
Changes done:
- include the app in the web image along with apache: this will permit to deploy the image to a docker registry, and users shall be able to run kafka without checking out sources
- add redis to compose file
- WARNING: comment out `php artisan package:discover` in composer.json. This is because it requires a DB to run, but it is done during install step, when the DB is not available. This seems to be a bug in artisan.
- remove all vagrant stuff
- update README.md and lint it (with markdown lint vs code extension)

Note there is no trello ticket cause I have only read access.


composer.json
```diff
@@ -86,7 +86,7 @@
         ],
         "post-autoload-dump": [
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
-            "@php artisan package:discover"
+            "#@php artisan package:discover"
```
ilestis, line 89: I don't think we can comment out this line. It is required by laravel to know where to find plugins. Removing it will mean adding new packages won't be auto-loaded by Laravel, as far as I know.

In the latest versions of laravel, they have an `--ansi` flag, which might help running it in docker?
https://github.com/laravel/laravel/blob/master/composer.json

ilestis (changes requested):
It might make more sense to move all the Docker and Vagrant stuff out of the readme.md and into separate markdown files with links to them, to avoid spamming the readme with stuff that doesn't impact a lot of people?

composer.json
ilestis: Apparently if you disable the db completely, it will ignore that part. https://www.reddit.com/r/laravel/comments/dcc49y/laravel_without_database_connection/

graham2071: From what I've seen is this is not a docker issue. `php artisan` needs a database, otherwise it raises a `connection refused`. I had the same problem with a build on my local machine. Without a runnning database, this step always fails with a connection error.
After some digging, here what I understand (i know nothing about Laravel, sorry my nood understanding):
- `composer install` : composer fetches all packages liste in composer.json
- `php artisan package:discover`: helps laravel to discover any new supported libraries or packages when running `composer update` (source: https://stackoverflow.com/questions/54648518/what-the-meaning-of-discovered-package)
- `php artisan` commands always require a running database
- post-autoload-dump: the script is run during install and update.

So my understanding is that running `php artisan package:discover` in `post-autoload-dump` is not ok because it runs during `install`, while it is only useful for `update`. We can use the script `pre-update-cmd` to do that.

Ok with that ?

graham2071: Hey, I had not seen your second comment. You put me on the right track. Changing the connection to an in-memory sqlite during build seems to do the trick. I will update the PR soon.

ilestis (approved):

ilestis:
Awesome work!

PR #404, opened 2022-05-31
Work in Progress - API: User Roles

pgmillon:
https://trello.com/c/4xdQHFeA

### DIFF
--- a/app/Http/Controllers/Api/v1/CampaignMemberApiController.php
+++ b/app/Http/Controllers/Api/v1/CampaignMemberApiController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Api\v1;
 
 use App\Http\Resources\CampaignUserResource;
 use App\Models\Campaign;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/v1/CampaignMemberApiController.php
```diff
@@ -0,0 +1,24 @@
...
+
+        return CampaignUserResource::collection($campaign->members);
+    }
+
+    public function show(Campaign $campaign, CampaignUser $campaignUser)
+    {
+        $this->authorize('access', $campaign);
+
+        return new CampaignUserResource($campaignUser);
+    }
```
pgmillon, line 23: Hey, before going any further I wanted to check that I could correctly fetch the CampaignUser but while using this I only have an empty resource:
```
{
    "data": {
        "id": null,
        "roles": [],
        "user": null
    }
}
```
I'm probably missing something since this is a Pivot resource.
Any suggestions ?

PR #406, opened 2022-06-01
Implemented another account deletion check

spitfire305:
Website now ask user to write goodbye as a final check to make sure its intentional, before allowing them to permanently delete their account.

### DIFF
--- a/app/Rules/GoodBye.php
+++ b/app/Rules/GoodBye.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace App\Rules;
+
+use Illuminate\Contracts\Validation\Rule;
+use Illuminate\Support\Str;
+
+class GoodBye implements Rule
+{
+    /**
+     * Create a new rule instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
+
+    /**
+     * Determine if the validation rule passes.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value)
+    {
+        return Str::is($value, 'goodbye') or Str::is($value, 'Goodbye');
+    }
+
+    /**
+     * Get the validation error message.
+     *
+     * @return string
+     */
+    public function message()
+    {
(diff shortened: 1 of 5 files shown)


app/Rules/GoodBye.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * Get the validation error message.
+     *
+     * @return string
+     */
+    public function message()
+    {
+        return ('You must write goodbye to confirm the deletion of your account');
```
ilestis, line 39: The error message should be translatable by our translators. Use `__('validation.goodbye')`, and add this message in `resources/land/en/validation.php`.

Also, `goodbye` should be a [parameter](https://laravel.com/docs/8.x/localization#replacing-parameters-in-translation-strings) so that our translators don't translate it. So in `validation.php`

```php
'goodbye' => 'You must write :code to confirm the deletion of your account',
```

And in `GoodBye.php`
```php
__('validation.goodbye', ['code' => 'goodbye'])
```

resources/views/settings/account.blade.php
```diff
@@ -131,9 +131,15 @@
...
                         {{ __('profiles.sections.delete.warning') }}
                     </p>
+                    <p class="mt-3">
+                        {{ __('profiles.sections.delete.goodbye') }}
+                    </p>
 
                     <div class="py-5">
                         {!! Form::model($user, ['method' => 'PATCH', 'route' => ['settings.account.destroy']]) !!}
+                        <div class="form-group required">
+                        {!! Form::text('goodbye',null, ['class' => 'form-control']) !!}
```
ilestis, line 141: You can add in the `[]` array `['class' => 'form-control', 'required']`. This is an html property which will stop the form from being submitted as long as a required field isn't filled out :)

ilestis (changes requested):
Great first PR! A few tweaks are needed to make it perfect :)

app/Rules/GoodBye.php
spitfire305: Changes made!

PR #413, opened 2022-06-06
Attribute filters: filter empty value instances for an attribute

spitfire305:


app/Models/Concerns/Filterable.php
```diff
@@ -161,12 +161,19 @@ public function scopeFilter(Builder $query, $params)
                             })
                             ->where('att.name', $filterValue);
 
-
                         $attributeValue = Arr::get($params, 'attribute_value');
-                        if ($attributeValue !== '' && $attributeValue !== null) {
+                        if ($attributeValue !== '' && $attributeValue !== null && $attributeValue !=='!!') {
```
ilestis, line 165: Missing a space between !== and !! for PSR12 compatibility

app/Models/Concerns/Filterable.php
```diff
@@ -161,12 +161,19 @@ public function scopeFilter(Builder $query, $params)
...
                             $query
                             ->where('att.value', $attributeValue);
                         }
+
+                        if ($attributeValue == '!!') {
```
ilestis, line 170: Please use "===", this is for PHP to confirm the content and also the variable type (in this case a string) :)

ilestis (changes requested):
Tiny coding standards tweaks

PR #414, opened 2022-06-07
Support for searching multiple races in characters

spitfire305:


app/Models/Concerns/Filterable.php
```diff
@@ -315,7 +343,7 @@ protected function extractSearchOperator($value, string $key): array
     {
         $operator = 'like';
         $filterValue = $value;
-        if ($key !== 'tags') {
+        if (($key !== 'tags') && ($key !== 'races')) {
```
ilestis, line 346: Is this change actually needed?

ilestis (changes requested):
One tweak for code readibility, and another to confirm a change is needed

app/Models/Concerns/Filterable.php
spitfire305: Yes, if that change is not made the following error message is displayed when trying to search using races as a parameter:

`strncmp(): Argument #1 ($string1) must be of type string, array given`

app/Models/Concerns/Filterable.php
```diff
@@ -125,6 +125,34 @@ public function scopeFilter(Builder $query, $params)
...
+                        }
+                        $query = $this->joinEntity($query);
+
+                        // Make sure we always have an array
+                        if (!is_array($value)) {
+                            $value = [$value];
+                        }
+
+                        if (!empty($filterOption) && $filterOption == 'exclude') {
+                            $tagIds = [];
```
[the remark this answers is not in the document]
spitfire305: It seems I forgot to rename the variables, sorry!

PR #417, opened 2022-06-08
Worldbuilding prompt: list all participants and hide invalid entries.

spitfire305:


resources/views/front/community-events/_results.blade.php
```diff
@@ -25,6 +25,7 @@
     </thead>
 <tbody>
 @foreach ($model->rankedResults as $entry)
+    @if ($entry->rank !== -1)
```
ilestis, line 28: This check should be done in the query `->where('rank', '<>', -1)`, to keep the html clean of business logic

ilestis (changes requested):
Move logic from html to query

PR #421, opened 2022-06-10
A toast warning now pops out

spitfire305:
A toast warning now pops out when an invite link is copied to the clipboard from the members section of a campaign.


ilestis (approved):
Easy peazy, nice!

PR #425, opened 2022-06-10
Entity logs are now automatically deleted a year after their creation.

spitfire305:


app/Console/Commands/CleanupEntityLogs.php
```diff
@@ -61,6 +62,18 @@ public function handle()
             });
 
         $this->info('Cleaned up ' . $this->count . ' entity logs.');
+
+        DB::beginTransaction();
+        try {
+            $this->count = EntityLog::
+                where('updated_at', '<=', Carbon::now()->subDays($delay)->toDateString())
```
ilestis, line 69: Tiny detail, but `created_at` is more correct, as `updated_at` will be changed on superboosted campaigns by the first loop of this command.

ilestis (changes requested):
Please use created_at rather than updated_at

ilestis (approved):

PR #426, opened 2022-06-13
Order by followers

spitfire305:


resources/views/campaigns/show.blade.php
```diff
@@ -53,7 +53,7 @@
 
                                 @if ($campaign->isPublic())
                                 <dt>{{ __('campaigns.fields.followers') }}</dt>
-                                <dd>{{ \App\Facades\CampaignCache::followerCount() }}</dd>
```
ilestis, line 56: Check if CampaignCache::followerCount() is used anywhere else, and if not, you can delete the code associated to it in `app\Services\Caches\CampaignCacheService`. Clear up both `followerCount()` and `clearFollowerCount()`

app/Services/CampaignFollowService.php
```diff
@@ -47,6 +47,8 @@ public function remove(Campaign $campaign, User $user): bool
         if (empty($follow)) {
             return false;
         }
+        $campaign->follower--;
```
ilestis, line 50: This is a good place to have this code, however if someone creates or deletes a follow through the API (not possible yet, but part of a future ticket), CampaignFollowService isn't called, so this number won't be updated. Which means it's cleaner to add it in the `app\Observers\CampaignFollowerObserver::deleted()` method, so that we aren't linked to calling the service in the future.

app/Services/CampaignFollowService.php
```diff
@@ -61,6 +63,8 @@ public function add(Campaign $campaign, User $user): bool
         $follow = new CampaignFollower();
         $follow->campaign_id = $campaign->id;
         $follow->user_id = $user->id;
+        $campaign->follower++;
```
ilestis, line 66: Same thing as for deleting, we're better off adding this logic in `app\Observers\CampaignFollowerObserver::created` method. Check out https://laravel.com/docs/9.x/eloquent#observers on how we can use some Laravel goodness in this case :)

ilestis (changes requested):
Great job! A bit of refactoring to make it easier for us when we adding followers through the API

app/Observers/CampaignFollowerObserver.php
```diff
@@ -17,6 +17,9 @@ public function created(CampaignFollower $campaignFollower)
     {
         UserCache::clearFollows();
         CampaignCache::campaign($campaignFollower->campaign)->clearFollowerCount();
```
ilestis, line 19: This can be removed, we don't need the cache anymore ;)

app/Observers/CampaignFollowerObserver.php
```diff
@@ -26,5 +29,7 @@ public function deleted(CampaignFollower $campaignFollower)
...
```
ilestis, line 31: Same here, no more caching of this value in Redis

ilestis (changes requested):
We don't need the caching anymore since it's not in the db, so we can remove the calls + code related to the old cache :)

ilestis (approved):

PR #427, opened 2022-06-14
Its now possible to ban users.

spitfire305:


resources/views/campaigns/_users.blade.php
```diff
@@ -81,15 +84,17 @@
...
-                                            <li>
-                                                <a href="{{ route('identity.switch', $relation) }}" title="{{ __('campaigns.members.helpers.switch') }}" data-toggle="tooltip" class="switch-user">
-                                                    <i class="fa-solid fa-sign-in-alt" aria-hidden="true"></i>
-                                                    {{ __('campaigns.members.actions.switch') }}
-                                                </a>
-                                            </li>
-                                            <li class="divider"></li>
-                                        @endcan
+                                        @if (!$relation->user->isBanned())
+                                            @can('switch', $relation)
```
ilestis, line 88: Instead of doing an if here (which means forgetting to add it to the controller), we can add the check directly in the CampaignMemberPolicy and revert back the dropdown option

app/Http/Middleware/CheckIfUserBanned.php
```diff
@@ -0,0 +1,36 @@
...
+
+        if (auth()->check() && auth()->user()->banned_until && auth()->user()->banned_until->isFuture()){
+
+            if (auth()->user()->banned_until < Carbon::now()->addDays(7)){
+                $days = auth()->user()->banned_until->diffInDays(Carbon::now()->addDays(7));
+                auth()->logout();
+                return redirect()->route('login')->withErrors(trans_choice('crud.errors.temp_ban',$days,['days' => $days]));
+            }
+            auth()->logout();
+            return redirect()->route('login')->withErrors(__('crud.errors.banned'));
```
ilestis, line 31: The `crud` translation file is for general actions in Kanka. We already have a whole translation file for authentication, so please use that instead. `__('auth.banned.permanent', ....)` and `__('auth.banned.temporary', ....)`. Also please make the translation more "personal" for the user. "You've been banned for X days/You've been permanently banned."

app/Http/Middleware/CheckIfUserBanned.php
```diff
@@ -0,0 +1,36 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure(\Illuminate\Http\Request): (\Illuminate\Http\Response|\Illuminate\Http\RedirectResponse)  $next
+     * @return \Illuminate\Http\Response|\Illuminate\Http\RedirectResponse
+     */
+    public function handle(Request $request, Closure $next)
+    {
+
+        if (auth()->check() && auth()->user()->banned_until && auth()->user()->banned_until->isFuture()){
```
ilestis, line 23: You've added the `isBanned()` method on the User class, might as well use it :)

ilestis (changes requested):
Small improvements needed :)

ilestis (approved):
🍻

PR #428, opened 2022-06-15
Added the ability to bulk delete relations

spitfire305:
and their respective mirrored relations.


ilestis (changes requested):
Change view path and refactor the code for simplicity

ilestis (approved):
👏🏻  perfecto

PR #429, opened 2022-06-16
Users can now add eras from the timeline view.

spitfire305:


resources/views/timelines/_timeline.blade.php
```diff
@@ -94,11 +94,32 @@ class="btn btn-box-tool" role="button"
...
+        <a href="{{ route('timelines.timeline_eras.create', ['timeline' => $model, 'from' => 'view']) }}" class="btn btn-warning btn-sm"
+            data-url="{{ route('timelines.timeline_eras.create', ['timeline' => $model]) }}"
+            data-backdrop="static">
+            <i class="fa-solid fa-plus"></i> {{ __('timelines/eras.actions.add') }}
+        </a>
+        @endcan
     </div>
 @endforelse
-
+@if($position)
```
ilestis, line 112: We can skip using a new variable by simply checking `!$timeline->eras->isEmpty()`. Also, don't forget space between conditions and the first ( :)

resources/views/timelines/_timeline.blade.php
```diff
@@ -94,11 +94,32 @@ class="btn btn-box-tool" role="button"
...
     </ul>
 @empty
+    @php
+    $position = 0;
+    @endphp
     <div class="alert alert-warning">
-        <p>{{ __('timelines.helpers.no_era') }}</p>
+        <p>{{ __('timelines.helpers.no_era') }} </p>
+        @can('update', $timeline)
+        <br>
```
ilestis, line 103: Instead of an empty br (looks ugly when copy pasting), add the "mb-2" class to the p above. This will add a margin-bottom of 0.5rem

resources/views/timelines/_timeline.blade.php
```diff
@@ -94,11 +94,32 @@ class="btn btn-box-tool" role="button"
...
+        </a>
+        @endcan
     </div>
 @endforelse
-
+@if($position)
+    @can('update', $timeline)
+        <div class="text-center mb-5">
+            <a href="{{ route('timelines.timeline_eras.create', ['timeline' => $model, 'from' => 'view']) }}" class="btn btn-primary btn-sm"
+                data-url="{{ route('timelines.timeline_eras.create', ['timeline' => $model]) }}"
```
ilestis, line 116: No need for data-url and backdrop, they are only needed if the link opens in a modal :)

ilestis (changes requested):
Small tweaks

ilestis (approved):

PR #430, opened 2022-06-16
Added new role "Wordsmith", if the user has achievements the achievem…

spitfire305:
…ents block now shows on their profile with the badges to their corresponding achievements.

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -540,14 +540,14 @@ class User extends \Illuminate\Foundation\Auth\User
      */
     public function hasAchievements(): bool
     {
-        return $this->hasRole('wordsmith');
+        return $this->isWordsmith();
     }
 
     /**
      * Determine if a user has the Wordsmith role
      * @return bool
      */
-    public function hasWordsmith(): bool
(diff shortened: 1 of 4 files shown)


app/User.php
```diff
@@ -540,6 +540,15 @@ public function isBanned(): bool
      */
     public function hasAchievements(): bool
     {
-        return false;
+        return $this->hasRole('wordsmith');
```
ilestis, line 543: Why not use the new hasWordsmith role?

app/User.php
```diff
@@ -540,6 +540,15 @@ public function isBanned(): bool
...
+    }
+
+    /**
+     * Determine if a user has the Wordsmith role
+     * @return bool
+     */
+    public function hasWordsmith(): bool
```
ilestis, line 550: Since a user is part of a role, would be better named isWordsmith :)

ilestis (changes requested):
Tiny bit of refactoring needed that will make the code easier to read

ilestis (approved):

PR #431, opened 2022-06-17
Entity images are now exposed as CSS variables

spitfire305:


resources/views/entities/components/header_grid.blade.php
```diff
@@ -287,3 +288,15 @@
...
     </dialog>
 @endsection
+
+
+@section('styles')
+    @parent
+    <style>
+        /** Entity Images URL**/
+        :root {
+            --entity-image-url: {{$imageUrl}};
```
ilestis, line 298: Since these are strings, they need to be surrounded by ', so '{{ $imageUrl }}` (don't forget space either)

ilestis (changes requested):
Tiny change

ilestis (approved):

PR #432, opened 2022-06-17
Login page now offers a list of the 5 most recent logged in users as …

spitfire305:
…links that when clicked log in the admin as the clicked user.

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -13,7 +13,15 @@
 
     <form method="POST" action="{{ route('login') }}">
         {{ csrf_field() }}
-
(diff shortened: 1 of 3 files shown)


resources/views/auth/login.blade.php
```diff
@@ -13,7 +13,15 @@
 
     <form method="POST" action="{{ route('login') }}">
         {{ csrf_field() }}
-
+        <div>
```
ilestis, line 16: This should be a <ul> to be valid HTML :)

ilestis (changes requested):

ilestis (approved):

PR #434, opened 2022-06-20
Corrected a bug that prevented users to adding images to an era on su…

spitfire305:
…mmernote.


ilestis (approved):
An easy fix, wrapped in complexity 🤖

PR #436, opened 2022-06-21
Relations: bulk edit for mirrors.

spitfire305:


app/Policies/RelationPolicy.php
```diff
@@ -46,6 +46,11 @@ public function create(User $user)
      */
     public function update(User $user, Relation $relation)
     {
+        // If the relation is empty, this call is comming from the bulk delete check
```
ilestis, line 49: Not sure why this is needed? Updating already works. This code was added for delete where the model is empty

app/Services/BulkService.php
```diff
@@ -405,7 +405,11 @@ protected function updateRelations(array $filledFields)
             if ($relation->owner_id == Arr::get($filledFields, 'target_id') || ($relation->target_id == Arr::get($filledFields, 'owner_id'))) {
                 continue;
             }
-
+            if (request()->update_mirrored && $relation->mirror) {
+                $relation->mirror->update($filledFields);
+                $this->count++;
+                $this->total++;
```
ilestis, line 411: This will be a problem if you bulk update the target and push the changes to the mirror (as the mirror is now attached to itself from itself, causing a buggy loop).
The $filledFields should exclude `target_id` and `owner_id`, to avoid relations from one entity to itself :)

The count and total should only increment if there was something to update (`update()` returns true)

ilestis (changes requested):
Question & edge case handling

app/Services/BulkService.php
```diff
@@ -405,11 +405,14 @@ protected function updateRelations(array $filledFields)
             if ($relation->owner_id == Arr::get($filledFields, 'target_id') || ($relation->target_id == Arr::get($filledFields, 'owner_id'))) {
                 continue;
             }
-
+            if (request()->update_mirrored && $relation->mirror) {
+                $relation->mirror->update(\array_diff_key($filledFields, ["target_id" => "xy", "owner_id" => "xy"]));
```
ilestis, line 409: This is a lot of code on a single line, and not sure why VSCode added a \ in front of array_diff_key, which is a default php function.

Cleaner would be 
> use Illuminate\Support\Arr;
> ....
> $mirrorFields = Arr::except($filledFields, ['target_id', 'owner_id']);
> $relation->mirror->update($mirrorFields);

More info on https://laravel.com/docs/9.x/helpers#method-array-except

ilestis (changes requested):

ilestis (approved):
Nice!

PR #439, opened 2022-06-22
User images path

spitfire305:


app/Observers/UserObserver.php
```diff
@@ -37,7 +37,8 @@ public function saving(User $user)
         }
 
         // Handle image. Let's use a service for this.
-        ImageService::handle($user, $user->getTable(), 60, 'avatar');
+        $folderno = floor($user->id / 1000);
```
ilestis, line 40: If my userID is `135440`, the $folderno is `135.0` because a floor doesn't convert to an int. For that you need `(int) floor($user->id / 1000);` ;-)

ilestis (changes requested):
Needs int casting to avoid floating points

PR #441, opened 2022-06-23
A toast notification now shows when filters are copied to the clipboard.

spitfire305:


ilestis (approved):
Easiest ticket of the week :D

PR #443, opened 2022-06-27
If a boosted campaing goes unboosted the map markers now revert to th…

spitfire305:
…e default shape and the polygon markers stop appearing.

### DIFF
--- a/app/Models/MapMarker.php
+++ b/app/Models/MapMarker.php
@@ -519,7 +519,7 @@ class MapMarker extends Model
      */
     public function visible(): bool
     {
-        if (!$this->map->campaign->boosted() && $this->custom_shape) {
+        if ($this->isPolygon() && !$this->map->campaign->boosted()) {
             return false;
         }
         return empty($this->entity_id) || (!empty($this->entity) && !empty($this->entity->child));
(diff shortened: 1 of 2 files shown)


app/Models/MapMarker.php
```diff
@@ -519,6 +519,9 @@ public function backgroundColour(): string
      */
     public function visible(): bool
     {
+        if (!$this->map->campaign->boosted() && $this->custom_shape) {
```
ilestis, line 522: While this works, if you look at the database queries in the debugger, it's making 2 queries for every single marker, which would kill our database server on prod where some maps have >200 markers :)
This needs to be refactored to tell Laravel/Eloquent to only load the map and campaign once, since it's the same on every marker.

Edit `MapMarker.php`'s `markers()` function and add at the end
```
    ->with(['map', 'map.campaign'])
 ```
This is part of Eloquent's [Eager loading](https://laravel.com/docs/9.x/eloquent-relationships#eager-loading)

---

Or, use the CampaignLocalization facade in `visible()` and `markerIcon()`
```
$campaign = CampaignLocalization::getCampaign();
if ($this->custome_shape && !$campaign->boosted()) {
```

Same thing with the check on the custom_icon

ilestis (changes requested):
Performance issue on large maps

ilestis (changes requested):
Two last performance and code readability changes

ilestis (approved):

PR #446, opened 2022-06-29
Items: parent item

spitfire305:


resources/lang/en/items.php
```diff
@@ -12,6 +12,8 @@
         'price'     => 'Price',
         'size'      => 'Size',
         'type'      => 'Type',
+        'item_id'   => 'Parent Item ID',
```
ilestis, line 15: Our convention is just "Parent Item" for the parent field, "ID" is a too technical for our users :)

resources/lang/en/items.php
```diff
@@ -26,10 +28,14 @@
         'price'     => 'Price of the item',
         'size'      => 'Size, Weight, Dimensions',
         'type'      => 'Weapon, Potion, Artefact',
+        'parent'    => 'ID of the parent item'
```
ilestis, line 31: Same here, look what is used for the quest's parent field.

resources/lang/en/items.php
```diff
@@ -26,10 +28,14 @@
...
     ],
     'show'          => [
         'tabs'  => [
             'inventories'   => 'Inventories',
         ],
     ],
+    'hints'         => [
+        'items'    => 'A web of interlocking items can be built using the Parent Item field.',
```
ilestis, line 39: I can tell this was copied from quests, where it makes sense. However for items, we can simply use "Organise items by using the parent item field."

resources/views/items/form/_entry.blade.php
```diff
@@ -25,6 +25,14 @@
         @include('cruds.fields.character', ['quickCreator' => true])
     </div>
 </div>
+<div class="row">
```
ilestis, line 28: To be consistent with the rest of the app, the parent field should be placed higher up in the form. please place it after type.

app/Models/Item.php
```diff
@@ -79,6 +85,7 @@ class Item extends MiscModel
         'size',
         'location.name',
         'character.name',
+        'item_id',
     ];
```
ilestis, line 90: I think `protected $sortable` is missing here for the items subpage. If you have multiple sub items, the "order by" (clicking on the header field name) won't work without it.

app/Models/Item.php
```diff
@@ -182,7 +189,21 @@ public function inventories()
...
     }
-
+    /**
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function items()
+    {
+        return $this->hasMany('App\Models\Item', 'item_id', 'id');
+    }
+    /**
```
ilestis, line 199: Missing empty line

ilestis (changes requested):
Missing from the ticket: we also need the parent item visible in the items/datagrid.blade.php, and being able to order on it :)

resources/views/items/datagrid.blade.php
```diff
@@ -12,6 +12,16 @@
         // Name
         'name',
         'type',
+        //Item_id
+        [
+            'label' => __('items.fields.item_id'),
+            'field' => 'item_id',
+            'render' => function($model) {
+                if ($model->item_id) {                    
+                    return '<a href="' . route('quests.show', $model->item_id) . '">' . e($model->item->name) . '</a>';
```
ilestis, line 21: I don't think it should be `quests.show` ;)

resources/views/items/datagrid.blade.php
```diff
@@ -12,6 +12,16 @@
         // Name
         'name',
         'type',
+        //Item_id
+        [
+            'label' => __('items.fields.item_id'),
+            'field' => 'item_id',
+            'render' => function($model) {
+                if ($model->item_id) {
```
ilestis, line 20: If you test just on `item_id`, it won't test the permission engine for non-admin viewers of a campaign. You always need to test `model->item`, because if the user doesn't have access to the item, it will return null.

resources/views/organisations/datagrid.blade.php
```diff
@@ -17,10 +17,13 @@
         ],
         // Location
         [
-            'type' => 'avatar',
-            'parent' => 'location',
-            'parent_route' => 'locations',
-            'visible' => $campaignService->enabled('locations'),
+            'label' => __('quests.fields.quest'),
```
ilestis, line 20: I'm not sure why organisations were changed? Organisations don't have a parent quest

ilestis (changes requested):
Small fixes

resources/views/organisations/datagrid.blade.php
spitfire305: It was a mistake, its been reverted.

ilestis (approved):

PR #448, opened 2022-06-30
Users now have the option to save+edit on a dropdown menu when editin…

spitfire305:
…g a campaign theme

### DIFF
--- a/app/Http/Controllers/Campaign/StyleController.php
+++ b/app/Http/Controllers/Campaign/StyleController.php
@@ -112,6 +112,11 @@ class StyleController extends Controller
         $campaignStyle->update($request->only('name', 'content', 'is_enabled'));
         CampaignCache::clearStyles();
 
+        if ($request->has('submit-update')) {
+            return redirect()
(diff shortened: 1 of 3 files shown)


ilestis (changes requested):
Missing the same logic in the `store()` and some funky tab spacing

app/Http/Controllers/Campaign/StyleController.php
```diff
@@ -90,6 +90,11 @@ public function store(StoreCampaignStyle $request)
         $style->save();
         CampaignCache::clearStyles();
 
+        if ($request->has('submit-update')) {
+            return redirect()
+            ->route('campaign_styles.edit', [$style])
+            ->with('success', __('campaigns/styles.update.success', ['name' => $style->name]));
```
ilestis, line 96: Wrong "success" message ;-) should be `.create.`

ilestis (changes requested):

ilestis (approved):

PR #452, opened 2022-07-05
Added the ability to create an entity child from a dropdown menu.

spitfire305:


resources/views/entities/components/header.blade.php
```diff
@@ -147,6 +147,13 @@
                                         <i class="fa-solid fa-plus" aria-hidden="true"></i> {{ __('crud.actions.new') }}
                                     </a>
                                 </li>
+                                @if(Schema::hasColumn($entity->pluralType(), $entity->Type() . '_id'))
```
ilestis, line 150: We can avoid a DB check by checking on the model's attributes if it has a `parent_id` field. It's faster to check in PHP than to ask mysql each time :)

ilestis (changes requested):
Refactoring needed for better performance

ilestis (approved):

PR #453, opened 2022-07-06
Tags are now automatically applied on creation of a new entity.

spitfire305:


resources/views/components/form/tags.blade.php
```diff
@@ -40,7 +40,15 @@
             }
         }
     }
+} elseif (empty($model)) {
+    $tags = \App\Models\Tag::autoApplied()->get();
```
ilestis, line 44: For performance reason, we need to add `->with('entity')` before `->get()`, so that getting the entity object asosciated with each tag is only a single database. Why do we need to check the tag's entity? It's (always) because of the permission engine doing weird things :P

resources/views/tags/_tree.blade.php
```diff
@@ -48,6 +48,17 @@
             },
             'disableSort' => true,
         ],
+        [
+            'label' => '<i class="fa-solid fa-tag" title="' . __('tags.fields.is_auto_applied') . '"></i>',
```
ilestis, line 52: Missing `data-toggle="tooltip"` on both <i> elements, so that hovering it with the mouse will directly show a tooltip in javascript, rather than the browser's default one.

resources/views/tags/_tree.blade.php
```diff
@@ -48,6 +48,17 @@
...
+            'field' => 'is_auto_applied',
+            'render' => function($model) {
+                if ($model->isAutoApplied()) {
+                    return '<i class="fa-solid fa-tag" title="' . __('tags.fields.is_auto_applied') . '"></i>';
```
ilestis, line 56: Same here, missing `data-toggle="tooltip"`

resources/views/tags/datagrid.blade.php
```diff
@@ -47,6 +47,17 @@
             },
             'disableSort' => true,
         ],
+        [
+            'label' => '<i class="fa-solid fa-tag" title="' . __('tags.fields.is_auto_applied') . '"></i>',
```
ilestis, line 51: Same story regarding the tooltip :)

app/Models/Tag.php
```diff
@@ -147,7 +151,7 @@ public function scopePreparedWith(Builder $query)
      */
     public function datagridSelectFields(): array
     {
-        return ['tag_id', 'colour'];
+        return ['tag_id', 'colour', 'is_auto_applied'];
```
ilestis, line 154: 👍🏻  awesome !

app/Models/Tag.php
```diff
@@ -332,6 +336,14 @@ public function showProfileInfo(): bool
         return false;
     }
 
+    /**
```
ilestis, line 339: Would be great to have a small description of the function of what it's used for like
```
/**
 * Determine if a tag is auto-applied to newly created entities
 * @return bool
 */
```

app/Services/EntityService.php
```diff
@@ -657,7 +657,14 @@ public function makeNewMentionEntity(MiscModel $model, string $name)
         $model->forceSavedObserver = true;
         $model->is_private = $defaultPrivate;
         $model->save();
-
+        if ($model->entity->type() != 'tag') {
+            $tags = \App\Models\Tag::autoApplied()->get();
```
ilestis, line 661: Same thing here about `->with('entity')`

app/Services/EntityService.php
```diff
@@ -657,7 +657,14 @@ public function makeNewMentionEntity(MiscModel $model, string $name)
         $model->forceSavedObserver = true;
         $model->is_private = $defaultPrivate;
         $model->save();
-
+        if ($model->entity->type() != 'tag') {
```
ilestis, line 660: This works, but we have a slightly faster option that will compare numbers rather than strings.
`$model->entity->type_id !== config('entities.ids.tag')`

app/Services/EntityService.php
```diff
@@ -657,7 +657,14 @@ public function makeNewMentionEntity(MiscModel $model, string $name)
...
+            $tags = \App\Models\Tag::autoApplied()->get();
+            foreach ($tags as $tag) {
+                if ($tag && $tag->entity) {
+                    $model->entity->tags()->attach($tag->id);
```
ilestis, line 664: If we're creating 5 new entities with [new:xxx] syntax and have 5 tags that auto-apply, this will generate 25 extra queries! We can 
a) only get the list of valid tag ids once and save it in an array in the EntityService
b) add all the tags ad once `attach()` accepts an array :)

ilestis (changes requested):
Good job! Few tweaks and some performance improvements possible before it's ready for production

app/Services/EntityService.php
```diff
@@ -657,13 +656,15 @@ public function makeNewMentionEntity(MiscModel $model, string $name)
         $model->forceSavedObserver = true;
         $model->is_private = $defaultPrivate;
         $model->save();
-        if ($model->entity->type() != 'tag') {
-            $tags = \App\Models\Tag::autoApplied()->get();
+        if ($model->entity->type_id !== config('entities.ids.tag')) {
+            $allTags = [];
+            $tags = \App\Models\Tag::autoApplied()->with('entity')->get();
```
ilestis, line 661: Nice improvement, but this loop is still executed for every new mention (fetching the tags). So if I have 100 `[new:character:ABC]` in the text, it will execute 100 times `App\Models\Tags::get()`.

Can you think of a way to improve this by doing the request only once? Maybe saving the values in a protected variable on `EntityService`? If you need inspiration, check out `EntityService::newEntityTypes()`

ilestis (changes requested):
Can we take it one step further to reduce database queries? :)

ilestis (approved):
Awesome!

PR #455, opened 2022-07-07
Quick creator: Posts

spitfire305:


app/Http/Controllers/EntityCreatorController.php
```diff
@@ -76,8 +83,13 @@ public function form($type)
     public function store(Request $request, $type)
     {
         // Make sure the user is allowed to create this kind of entity
-        $class = $this->entityService->getClass($type);
-        $this->authorize('create', $class);
+        if ($type != 'posts') {
```
ilestis, line 86: Just for consistency, let's keep the same test context from in the `form()` method (checking == 'post' first)

app/Http/Controllers/EntityCreatorController.php
```diff
@@ -104,15 +116,24 @@ public function store(Request $request, $type)
...
+                $model = new $class;
+                $new = $model->create($values);
+                $new->crudSaved();
+                $new->entity->crudSaved();
+            } else {
+                if ($values['position'] == 0) {
+                    $new = EntityNote::create($values);
+                } else {
+                    $entity = Entity::find($values['entity_id']);
+                    $entity->notes()->increment('position');
```
ilestis, line 132: Ooooh didn't know about `->increment()`, that's very cool!

app/Http/Requests/StorePost.php
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'name' => 'required|max:191',
+            'entity_id' => 'required',
```
ilestis, line 28: Worth adding a `|numeric|exists:entities,id` to make sure it's a valid entity_id before saving to the database. Also worth validating `visibility_id` the same way it's validated in other rules.

app/Http/Controllers/EntityCreatorController.php
```diff
@@ -104,15 +116,24 @@ public function store(Request $request, $type)
...
+            if ($type != 'posts') {
+                $this->validateEntity($values, $validator->rules());
+
+                /** @var MiscModel $model */
+                $model = new $class;
+                $new = $model->create($values);
+                $new->crudSaved();
+                $new->entity->crudSaved();
+            } else {
+                if ($values['position'] == 0) {
```
ilestis, line 128: Worth adding a comment before the if to explain the difference between the if and else, so it's faster to parse :) I wasn't sure if the first if is for position first or position last without having to look deeper in the condition

resources/lang/en/posts.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+return [
+    'create'        => [
+        'title' => 'New Post',
+    ],
+    'fields'        => [
+        'name'          => 'Name',
+        'parent'        => 'Parent Entity',
```
ilestis, line 9: Parent entity in this case would be wrong, when compared to the terminology in the rest of the app. I use "parent" when it's the same nested. You can remove this key, and use `crud.fields.entity` in the form instead

resources/lang/en/posts.php
```diff
@@ -0,0 +1,19 @@
...
+        'position'      => 'Position',
```
ilestis, line 10: We already have this "position" field defined in `entities/notes.fields.position`, so remove it from this file and change the form

resources/lang/en/posts.php
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    'placeholders'  => [
+        'name'  => 'Name of the post',
```
ilestis, line 13: Same here, we have all these translations already in all languages of kanka in `entities/notes` translation file :)

resources/views/entities/creator/form.blade.php
```diff
@@ -37,14 +37,15 @@
...
+    @if ($type !== 'tags' && $type !== 'posts')
         <div id="quick-creator-tags-field">
     @include('cruds.fields.tags', ['dropdownParent' => '#quick-creator-tags-field'])
         </div>
     @endif
 
-    @includeWhen(auth()->user()->isAdmin(), 'cruds.fields.privacy_callout')
-
+    @if ($type !== 'posts')
+        @includeWhen(auth()->user()->isAdmin(), 'cruds.fields.privacy_callout')
```
ilestis, line 47: `@includeWhen()` evaluated the first parameter as a condition, so we can combine auth()->user()->isAdmin() && $type !== 'post', making the code easier to read :)

ilestis (changes requested):
Nice job! A bit of cleanup needed, and using existing translations (I know it's confusing that it's posts but the model and old code is called entity_notes :P)

resources/lang/en/posts.php
spitfire305: These are auto generated by the code.

PR #459, opened 2022-07-13
Fixed a bug where tags where auto applying to widgets and random enti…

spitfire305:
…ties when they should not.

### DIFF
--- a/resources/views/dashboard/widgets/forms/_random.blade.php
+++ b/resources/views/dashboard/widgets/forms/_random.blade.php
@@ -58,7 +58,8 @@
                                 'tag_id',
                                 [
                                     'model' => isset($model) ? $model : null,
-                                    'enableNew' => false
+                                    'enableNew' => false,
+                                    'enableAutoTags' => null,
                                 ]
                             ) !!}
                         <input type="hidden" name="save_tags" value="1" />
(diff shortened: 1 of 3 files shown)


resources/views/dashboard/widgets/forms/_random.blade.php
```diff
@@ -58,7 +58,8 @@
                                 'tag_id',
                                 [
                                     'model' => isset($model) ? $model : null,
-                                    'enableNew' => false
+                                    'enableNew' => false,
+                                    'enableAutoTags' => null,
```
ilestis, line 62: I feel like `false` would be more correct and easier to read then `null`

resources/views/dashboard/widgets/forms/_tags.blade.php
```diff
@@ -4,7 +4,8 @@
             [
                 'model' => isset($model) ? $model : null,
                 'enableNew' => false,
-                'helper' => __('dashboard.widgets.recent.tags')
+                'helper' => __('dashboard.widgets.recent.tags'),
+                'enableAutoTags' => null,
```
ilestis, line 8: Same here, `false` makes more sense

ilestis (changes requested):
Please replace `null` with `false`

ilestis (approved):

PR #463, opened 2022-07-15
Users can now set entity files as pinned directly on the overview sec…

spitfire305:
…tion.

### DIFF
--- a/resources/views/entities/components/assets.blade.php
+++ b/resources/views/entities/components/assets.blade.php
@@ -1,0 +1,18 @@
+<?php
+/**
+ * @var \App\Models\MiscModel $model
+ * @var \App\Models\EntityAsset $asset
+ */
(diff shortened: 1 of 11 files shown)


resources/views/entities/components/assets.blade.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * @var \App\Models\MiscModel $model
+ * @var \App\Models\EntityAsset $asset
+ */
+$attributes = $model->entity->assets->where('is_pinned', 1)->where('type_id', 1);
```
ilestis, line 6: I think the variable could be better named ;-)

ilestis (changes requested):
Great job! One tiny change :)

ilestis (approved):

PR #467, opened 2022-07-18
Added the ability to set a reminder as the foundation date for organi…

spitfire305:
…sations, locations and families.

### DIFF
--- a/app/Models/EntityEvent.php
+++ b/app/Models/EntityEvent.php
@@ -654,7 +654,7 @@ class EntityEvent extends MiscModel
     }
 
     /**
-     * Determine if an event is of the character birth type
+     * Determine if an event is of the entity foundation type
      * @return bool
      */
     public function isFounded(): bool
--- a/app/Models/EntityEvent.php
+++ b/app/Models/EntityEvent.php
@@ -654,6 +654,15 @@ class EntityEvent extends MiscModel
     }
 
     /**
+     * Determine if an event is of the character birth type
+     * @return bool
+     */
+    public function isFounded(): bool
+    {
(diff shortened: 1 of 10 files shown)


app/Models/EntityEvent.php
```diff
@@ -653,6 +653,15 @@ public function isBirth(): bool
         return $this->type_id === EntityEventType::BIRTH;
     }
 
+    /**
+     * Determine if an event is of the character birth type
```
ilestis, line 657: I think the function's comment might not be true :)

resources/views/calendars/events/_subform.blade.php
```diff
@@ -82,3 +82,10 @@
         <p class="help-block">{!! __('entities/events.helpers.characters', ['more' => link_to_route('helpers.age', __('crud.actions.find_out_more'), null, ['target' => '_blank'])]) !!}</p>
     </div>
 @endif
+@if (!empty($entity) && $entity->typeId() == config('entities.ids.location') || $entity->typeId() == config('entities.ids.family') || $entity->typeId() == config('entities.ids.organisation'))
```
ilestis, line 85: `$entity->typeId() == config('entities.ids.location') || $entity->typeId() == config('entities.ids.family') || $entity->typeId() == config('entities.ids.organisation')` can be replaced with an `in_array($entity->typeId(), [...])`

ilestis (changes requested):
Small tweaks and a question: if a location only has a name, entry, and a foundation, does it appear in the profile sidebar? I don't think it would, because location (and family and org) are missing that check in the model's `showProfile()`

ilestis (approved):

PR #469, opened 2022-07-20
Added the ability to replace entity mentions for the new entities cre…

spitfire305:
…ated when copiying an entity.

### DIFF
--- a/app/Observers/EntityObserver.php
+++ b/app/Observers/EntityObserver.php
@@ -117,7 +117,6 @@ class EntityObserver
         } elseif (request()->has('copy_source_permissions') && request()->filled('copy_source_permissions')) {
             return;
         }
-
         $data = request()->only('role', 'user', 'is_attributes_private', 'permissions_too_many');
 
         // If the user granted/assigned themselves read/write permissions on the entity, we need to make sure they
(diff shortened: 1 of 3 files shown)


app/Observers/EntityObserver.php
```diff
@@ -156,7 +155,32 @@ protected function saveAttributes(Entity $entity): self
             'is_attributes_private'
         );
         $this->attributeService->saveEntity($data, $entity);
-
+        $sourceId = request()->post('copy_source_id');
+        if (request()->has('replace_mentions') && request()->filled('replace_mentions')) {
+            $source = $source ?? Entity::findOrFail($sourceId);
```
ilestis, line 160: I don't think $source was previously set so the condition will always be false

app/Observers/EntityObserver.php
```diff
@@ -156,7 +155,32 @@ protected function saveAttributes(Entity $entity): self
...
+            $sourceAttributes = [];
+            $entityAttributes = [];
+            foreach ($source->attributes as $attribute) {
+                array_push($sourceAttributes, '{attribute:' . $attribute->id . '}');
+            }
+            foreach ($entity->attributes as $attribute) {
+                array_push($entityAttributes, '{attribute:' . $attribute->id . '}');
+            }
+            $attributes = array_combine($sourceAttributes, $entityAttributes);
```
ilestis, line 169: This whole block can be replaced by
```$entry = Str::replace($sourceAttributes, $entityAttributes, $entity->child->entry);```, using Laravel's [Str helper](https://laravel.com/docs/9.x/helpers#method-str-replace)

File: app/Observers/EntityObserver.php [hunk 3]
```diff
@@ -156,7 +155,32 @@ protected function saveAttributes(Entity $entity): self
...
+            $entry = $entity->child->entry;
+            foreach ($attributes as $sourceAttribute => $entityAttribute) {
+                $entry = str_replace($sourceAttribute, $entityAttribute, $entry);
+            }
+            $entity->child->update(['entry' => $entry]);
+            foreach ($entity->notes as $note) {
+                $entry = $note->entry;
+                foreach ($attributes as $sourceAttribute => $entityAttribute) {
```
ilestis, line 177: Same thing, the foreach can be replaced by `Str::replace()`

ilestis (changes requested):
Awesome job figuring out this tricky ticket! Some small optimizations are possible

ilestis (approved):

PR #471, opened 2022-07-22
Added CleanupJobLogs command.

spitfire305:


app/Console/Commands/CleanupJobLogs.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
ilestis, line 1: This job would make more sense in the admin project :)

app/Console/Commands/CampaignVisibileEntityCount.php
```diff
@@ -57,8 +58,14 @@ public function handle()
                 $campaign->save();
             }
         });
-
-        $this->info('Updated ' . $this->count . ' public campaigns.');
+        $log = "Updated {$this->count} public campaigns.";
+        $this->info($log);
+        if (config('app.log_jobs')) {
+            JobLog::create([
+                'name' => 'campaign-visible-entity-count',
```
ilestis, line 65: Can you everywhere replace `'name' => '...'` with `'name' => $this->signature`? I think that will be easier to read and then look up in the code :)

app/Console/Kernel.php
```diff
@@ -40,6 +41,7 @@ protected function schedule(Schedule $schedule)
         $schedule->command(CleanupEntityLogs::class)->dailyAt('03:30');
         $schedule->command(CleanupEntityUsers::class)->dailyAt('03:35');
         $schedule->command(CleanupUserLog::class)->dailyAt('03:45');
+        $schedule->command(CleanupJobLog::class)->weeklyOn(1, '03:55');
```
ilestis, line 44: Please move this to the admin project's kernel :)

ilestis (changes requested):
Please move the cleanup job to the admin project, and I've changed my mind on the `'name'` value 🙈

ilestis (approved):
Perfect

PR #472, opened 2022-07-25
Added the ability for an user to impersonate another user inside an e…

spitfire305:
…ntity's overview page

### DIFF
--- a/app/Http/Controllers/Campaign/MemberController.php
+++ b/app/Http/Controllers/Campaign/MemberController.php
@@ -6,6 +6,7 @@ use App\Facades\Identity;
 use App\Http\Controllers\Controller;
 use App\Models\CampaignRole;
 use App\Models\CampaignUser;
+use App\Models\Entity;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Campaign/MemberController.php
```diff
@@ -29,11 +30,15 @@ public function __construct(MemberService $memberService)
      * Switch to a member of the campaign
      * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
      */
-    public function switch(CampaignUser $campaignUser)
+    public function switch(CampaignUser $campaignUser, Entity $entity)
```
ilestis, line 33: Please set the entity's default to null, so it's easier to read (it's an optional parameter) :)
`Entity $entity = null`

ilestis (changes requested):
Nice job! Tiny code quality of life change :)

ilestis (approved):

PR #474, opened 2022-08-02
Updating a relation now redirects to the correct page.

spitfire305:


app/Http/Controllers/Entity/RelationController.php
```diff
@@ -214,6 +213,9 @@ public function update(StoreRelation $request, Entity $entity, Relation $relatio
                 if (!empty($mode)) {
                     $redirect['mode'] = $mode;
                 }
+                if (request()->option) {
```
ilestis, line 216: I personally find `request()->has('option')` to be easier to read :)

app/Http/Controllers/Entity/RelationController.php
```diff
@@ -214,6 +213,9 @@ public function update(StoreRelation $request, Entity $entity, Relation $relatio
...
+                    $redirect['option'] = request()->option;
```
ilestis, line 217: Same here `request()->get('option')` make it clearer and appear less "magic", which is easier to read

app/Http/Controllers/Entity/RelationController.php
```diff
@@ -229,6 +231,9 @@ public function update(StoreRelation $request, Entity $entity, Relation $relatio
         if (!empty($mode)) {
             $redirect['mode'] = $mode;
         }
+        if (request()->option) {
```
ilestis, line 234: Same here with has() and get()

app/Http/Controllers/Entity/RelationController.php
```diff
@@ -246,6 +251,12 @@ public function destroy(Entity $entity, Relation $relation)
     {
         $this->authorize('update', $entity->child);
 
+        if (request()->mode) {
```
ilestis, line 254: Same here 👯

app/Http/Controllers/Entity/RelationController.php
```diff
@@ -264,9 +275,15 @@ public function destroy(Entity $entity, Relation $relation)
         }
 
         $relation->delete();
-
+        $redirect = [$entity];
+        if (!empty($mode)) {
+            $redirect['mode'] = $mode;
+        }
+        if (request()->option) {
```
ilestis, line 282: I sound like a broken clock at this point

resources/views/entities/pages/relations/update.blade.php
```diff
@@ -67,14 +67,16 @@
...
-    {!! Form::close() !!}
+    {!! Form::hidden('option', request()->option) !!}
+    {!! Form::hidden('mode', request()->mode) !!}
 
+    {!! Form::close() !!}
     {!! Form::open([
         'method' => 'DELETE',
-        'route' => ['entities.relations.destroy', $entity->id, $relation->id],
-        'id' => 'delete-relation-' . $relation->id]) !!}
+        'route' => ['entities.relations.destroy', 'entity' => $entity->id, 'relation' => $relation->id, 'mode' => request()->mode, 'option' => request()->option],
```
ilestis, line 77: Same here, 2x`get()` means that in the future we can also give a second, fallback value if we wish to build one in

ilestis (changes requested):
Works, but I'm picky with making the code easy to read :D

ilestis (approved):
Nice!

PR #475, opened 2022-08-03
Corrected minor bug.

spitfire305:


ilestis (approved):
I was about to write how the logic is wrong, but no you are correct 😂

PR #476, opened 2022-08-04
Added the ability for users to select the default option for the rela…

spitfire305:
…tion view.

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -373,6 +373,14 @@ class Campaign extends MiscModel
     /**
      * @return mixed
      */
+    public function defaultToConnectionMode(): int
+    {
+        return (int) Arr::get($this->ui_settings, 'connections_mode', false);
+    }
+
+    /**
+     * @return mixed
+     */
     public function getHideMembersAttribute()
     {
         return Arr::get($this->ui_settings, 'hide_members', false);
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -375,7 +375,7 @@ class Campaign extends MiscModel
      */
     public function defaultToConnectionMode(): int
     {
-        return (int) Arr::get($this->ui_settings, 'connections_mode', false);
+        return (int) Arr::get($this->ui_settings, 'connections_mode', 0);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/Models/Campaign.php
```diff
@@ -370,6 +370,14 @@ public function defaultToConnection(): bool
         return (bool) Arr::get($this->ui_settings, 'connections', false);
     }
 
+    /**
+     * @return mixed
+     */
+    public function defaultToConnectionMode(): int
```
ilestis, line 376: I feel like the fallback from Arr::get should be an int rather than a book then cast to an int, makes the code easier to read :)

resources/lang/en/campaigns.php
```diff
@@ -417,13 +417,20 @@
         'connections'       => [
             'explorer'  => 'Relations explorer (if available, for boosted campaigns)',
             'list'      => 'List interface',
+            'default'   => 'Default',
```
ilestis, line 420: Why not reuse the existing translations, so that our translators don't need to work twice? :)

resources/views/campaigns/forms/panes/ui.blade.php
```diff
@@ -121,6 +121,18 @@
...
 
+        <div class="col-md-6 col-lg-4">
+            <div class="form-group">
+                <label>
+                    {{ __('campaigns.ui.fields.connections_mode') }}
+                    <i class="fa-solid fa-question-circle hidden-xs hidden-sm" data-toggle="tooltip" title="{{ __('campaigns.ui.helpers.connections_mode') }}"></i>
+
+                </label>
+                {!! Form::select('ui_settings[connections_mode]', [0 => __('campaigns.ui.connections.default'), 1 => __('campaigns.ui.connections.direct'), 2 => __('campaigns.ui.connections.related'), 3 => __('campaigns.ui.connections.mentions')], null, ['class' => 'form-control']) !!}
+                <p class="help-block visible-xs visible-sm">{{ __('campaigns.ui.helpers.connections') }}</p>
```
ilestis, line 132: This second block is for mobile view, and it doest seem to use the same translation key than the tooltip. It should be the same key, since on mobile we don't have tooltips

ilestis (changes requested):
Small changes :)

ilestis (approved):
🎉

PR #477, opened 2022-08-04
Application Aproval Message

spitfire305:


app/Services/Campaign/SubmissionService.php
```diff
@@ -66,7 +66,7 @@ public function process(array $data): string
                         'reason' => $rejection
                     ]));
         } else {
-            $this->approve((int) Arr::get($data, 'role_id'));
+            $this->approve((int) Arr::get($data, 'role_id'), $data);
```
ilestis, line 69: This code looks weird, the first Param is getting the role is from the data, but the second Param is the whole data. Can we either only use data, or call the method with the two params we need?

resources/lang/en/notifications.php
```diff
@@ -3,7 +3,7 @@
 return [
     'campaign'          => [
         'application'           => [
-            'approved'  => 'Your application to the :campaign campaign has been approved.',
+            'approved'  => 'Your application to the :campaign campaign has been approved. Message provided: :reason',
```
ilestis, line 6: I know the rejection shows an empty reason, which isn't great UX. Can we have two strings, one with and one without the reason, and the notification is saved with the correct one? Thanks :)

ilestis (changes requested):
Small changes :)

resources/lang/en/notifications.php
```diff
@@ -3,9 +3,12 @@
...
```
ilestis, line 6: By replacing the base key and adding a new key for the old translation, our translators won't know that the key needs to be updated. So to make their lives easier, please use the same key for approval, and create a new one for if there is a reason message :)

ilestis (changes requested):
Leaving the existing translations as is to make our translator lives easier

ilestis (approved):
Nice, thanks!

PR #478, opened 2022-08-09
Added the ability for the user to view and assign permissions from th…

spitfire305:
…e api

### DIFF
--- a/app/Http/Resources/EntityPermissionResource.php
+++ b/app/Http/Resources/EntityPermissionResource.php
@@ -14,11 +14,13 @@ class EntityPermissionResource extends EntityChild
      */
     public function toArray($request)
     {
-        return $this->entity([
+        return [
             'role' => $this->campaign_role_id,
             'user' => $this->user_id,
             'action' => $this->action,
-            'access' => $this->access,
-        ]);
+            'access' => (bool) $this->access,
+            'created_at' => $this->created_at,
+            'updated_at' => $this->updated_at,
+        ];
     }
 }
--- a/app/Http/Resources/EntityPermissionResource.php
+++ b/app/Http/Resources/EntityPermissionResource.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Http\Resources;
+
+use Illuminate\Http\Resources\Json\JsonResource;
+
+class EntityPermissionResource extends EntityChild
+{
+    /**
+     * Transform the resource into an array.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return array
+     */
+    public function toArray($request)
+    {
+        return $this->entity([
+            'role' => $this->campaign_role_id,
+            'user' => $this->user_id,
+            'action' => $this->action,
+            'access' => $this->access,
+        ]);
+    }
+}
--- a/app/Http/Resources/EntityPermissionResource.php
+++ b/app/Http/Resources/EntityPermissionResource.php
@@ -15,8 +15,9 @@ class EntityPermissionResource extends EntityChild
     public function toArray($request)
     {
         return [
-            'role' => $this->campaign_role_id,
-            'user' => $this->user_id,
+            'id'   => $this->id,
+            'campaign_role_id' => $this->campaign_role_id,
+            'user_id' => $this->user_id,
             'action' => $this->action,
             'access' => (bool) $this->access,
             'created_at' => $this->created_at,
--- a/app/Services/Api/ApiPermissionService.php
+++ b/app/Services/Api/ApiPermissionService.php
@@ -11,34 +11,12 @@ use Illuminate\Support\Facades\DB;
 
 class ApiPermissionService
 {
-    /**
-     * Set the entity type
-     * @param string $type
-     * @return PermissionService
-     */
-    public function type(int $type): self
-    {
-        $this->type = $type;
-        return $this;
-    }
-
-    /**
-     * Set the role
-     * @param CampaignRole $role
-     * @return $this
-     */
-    public function role(CampaignRole $role): self
-    {
-        $this->role = $role;
-        return $this;
-    }
-
      /**
      * Get the permissions of an entity
      * @param Entity $entity
      * @return mixed
      */
-    public function entityPermissions(Entity $entity): array
+    protected function entityPermissions(Entity $entity): array
     {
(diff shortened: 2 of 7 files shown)


app/Services/Api/ApiPermissionService.php
```diff
@@ -0,0 +1,84 @@
...
+use Illuminate\Support\Facades\DB;
+
+class ApiPermissionService
+{
+    /**
+     * Set the entity type
+     * @param string $type
+     * @return PermissionService
+     */
+    public function type(int $type): self
```
ilestis, line 19: I don't see this being used anywhere

app/Services/Api/ApiPermissionService.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->type = $type;
+        return $this;
+    }
+
+    /**
+     * Set the role
+     * @param CampaignRole $role
+     * @return $this
+     */
+    public function role(CampaignRole $role): self
```
ilestis, line 30: I don't see this being used anywhere?

app/Services/Api/ApiPermissionService.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->role = $role;
+        return $this;
+    }
+
+     /**
+     * Get the permissions of an entity
+     * @param Entity $entity
+     * @return mixed
+     */
+    public function entityPermissions(Entity $entity): array
```
ilestis, line 41: Since this is only called in the service, it's worth making the function `protected` instead of `public`.

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Results
+```json
+{
+    "data": [
+        {
+            "access": 1,
+            "action": 1,
+            "created_at": "2022-07-25T16:21:38.000000Z",
+            "created_by": null,
```
ilestis, line 28: Once `$this->entity` is removed, this needs to be cleaned up too (no created_by, is_private, updated_by`)

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -0,0 +1,60 @@
...
+
+| Method | URI | Headers |
+| :- |   :-   |  :-  |
+| POST | `entities/{entity.id}/entity_permissions` | Default |
+
+### Body
+
+| Parameter | Type | Detail |
+| :- |   :-   |  :-  |
+| `campaign_role_id` | `integer` (Required) | The campaign role id affected by the permission, only required when theres no user id |
```
ilestis, line 52: Typo on "there's". Same on the next line

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -0,0 +1,60 @@
...
+| `user_id` | `integer` (Required) | The id of the user affected by the permission, only required when theres no campaign role id|
+| `action` | `integer` (Required) | The code of the action controller by the permission |
```
ilestis, line 54: The available actions are unclear to users. Please add a table below in the docs with a list of IDs and what they allow. You can find that info in `\App\Models\CampaignPermission` (only 1 to 6)

ilestis (changes requested):
Small changes, but awesome job on such a complex ticket!

app/Http/Resources/EntityPermissionResource.php
```diff
@@ -14,11 +14,13 @@ class EntityPermissionResource extends EntityChild
      */
     public function toArray($request)
     {
-        return $this->entity([
+        return [
             'role' => $this->campaign_role_id,
```
ilestis, line 18: Forgot to specify that the `id` also needs to be added, just to be safe :)

app/Services/Api/ApiPermissionService.php
```diff
@@ -62,6 +40,7 @@ public function saveEntity($request, Entity $entity)
     {
```
ilestis, line 40: The function can now type cast the return with `public function saveEntity(...) : CampaignPermission`, and the auto-docs of the function call updated.

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -22,17 +22,12 @@ You can get a list of all the entity-permissions of an entity by using the follo
...
         {
-            "access": 1,
+            "role": 115,
+            "user": null,
             "action": 1,
+            "access": true,
             "created_at": "2022-07-25T16:21:38.000000Z",
-            "created_by": null,
-            "entity_id": 273,
-            "id": 37,
```
ilestis: Still need ID, sorry about that :)

app/Http/Resources/EntityPermissionResource.php
```diff
@@ -14,11 +14,13 @@ class EntityPermissionResource extends EntityChild
...
             'user' => $this->user_id,
```
ilestis, line 19: Also to make it clearer for users, please use `user_id` and `campaign_role_id`. The closer the API is to the db, the easier it is for us to debug

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -22,17 +22,12 @@ You can get a list of all the entity-permissions of an entity by using the follo
 {
     "data": [
         {
-            "access": 1,
+            "role": 115,
```
ilestis, line 25: Replace with `campaign_role_id` and `user_id`

resources/api-docs/1.0/entity-permissions.md
```diff
@@ -49,12 +44,36 @@ To create an entity-tag, use the following endpoint.
...
 ### Results
 
-> {success} Code 200 with JSON.
+> {success} Array with all the newly created permissions.
+
+```json
+{
+    "data": [
+        {
+            "role": null,
```
ilestis, line 70: Change also here the examples to campaign_role_id, user_id, and add id.

ilestis (changes requested):
Last changes to the API and it's good to go :)

app/Services/Api/ApiPermissionService.php
spitfire305: It doesnt return a CampaignPermission. it returns an array containing all the CampaingPermissions created, if I do the type cast I get an error when trying to post on the api

ilestis (approved):

PR #479, opened 2022-08-10
Added the ability to filter entities by children location.

spitfire305:


app/Models/Concerns/HasFilters.php
```diff
@@ -421,7 +424,53 @@ protected function filterRaces(Builder $query, string $value = null): void
...
+    /**
+     * Filter on characters on multiple locations
+     * @param Builder $query
+     * @param string|null $value
+     * @return void
+     */
+    protected function filterLocations(Builder $query, string $value = null): void
+    {
+        $key = 'location_id';
+        if ($this->filterOption('exclude')) {
```
ilestis, line 436: A lot of this is duplicate code, is it not possible to call the fallback function if it's not `children`?

ilestis (changes requested):
Missing the new option on character races, families, and orgs. Will need to adapt the docs, since it won't be just locations. And avoid copy-pasting lots of code :P

resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -60,7 +60,7 @@
                                     @include('cruds.datagrids.filters._tag')
                                 @elseif ($field['type'] === 'select')
                                     @include('cruds.datagrids.filters._select')
-                                @elseif ($field['label'] === 'Location')
+                                @elseif ($field['label'] === 'Location' || $field['label'] === 'Family' || $field['label'] === 'Race' || $field['label'] === 'Organisation')
                                     @include('cruds.datagrids.filters._location')
```
ilestis, line 64: I feel like the `_location` file should be renamed, since it's not limited to just locations. I also feel  like we don't _need_ another file, but instead it could simply be an option from the way the Filter class is built. For example, in the `CharacterFilter` class, add an option `'withChildren' => true`, and in the `_select` file, simply check if the option is enabled ;-) that way we don't need to add a new if in this file each time we add the option to a filter.

app/Models/Concerns/HasFilters.php
```diff
@@ -117,7 +120,7 @@ public function scopeFilter(Builder $query, array $params = []): Builder
                     $this->filterDateRange($query, $key, $params);
                 } elseif ($key == 'races') {
                     $this->filterRaces($query, $value);
-                } elseif ($key == 'location_id') {
+                } elseif ($key == 'location_id' && $this->filterOption('children')) {
```
ilestis, line 123: We don't need the $this->filterOption here, we can make a fallback in the filterLocation function instead. Makes the code easier to read

app/Models/Concerns/HasFilters.php
```diff
@@ -432,44 +435,14 @@ protected function filterRaces(Builder $query, string $value = null): void
...
+        if ($this->filterOption('children')) {
             $location = Location::find($value);
             $locationIds = $location->descendants->pluck('id')->toArray();
             array_unshift($locationIds, $location->id);
-            foreach ($locationIds as $location) {
-                $query->orWhere($this->getTable() . '.location_id', '=', $location);
-            }
+            $query->whereIn($this->getTable() . '.location_id', $locationIds)->distinct();
+
             return;
```
ilestis, line 444: Missing `else { $this->defaultFilter() }`

app/Models/Concerns/HasFilters.php
```diff
@@ -432,44 +435,14 @@ protected function filterRaces(Builder $query, string $value = null): void
...
-                    $this->getTable() . '.' . $key,
-                    '!=',
-                    $this->filterValue
-                )->orWhereNull($this->getTable() . '.' . $key);
-            });
-
-            return;
-        } elseif ($this->filterOption('children')) {
+        if ($this->filterOption('children')) {
             $location = Location::find($value);
```
ilestis, line 439: If I call the page with `location_id=123123123`, the page crashes. Same with race, org, and family. Bots constantly send invalid data to forms, and it would spam me if we don't ignore invalid values ;-) You can simply do `if (empty($location)) { return }`

app/Models/Concerns/HasFilters.php
```diff
@@ -483,6 +456,18 @@ protected function filterRace(Builder $query, string $value = null): void
             $query->whereRaw('(select count(*) from character_race as cr where cr.character_id = ' .
                 $this->getTable() . '.id and cr.race_id = ' . ((int) $value) . ') = 0');
             return;
+        } elseif ($this->filterOption('children')) {
```
ilestis, line 459: There is a lot of duplicated code here between the children option and the default option. Can you think of a way of merging it together and avoid duplicate code? For races, families, and orgs.

ilestis (changes requested):
All works nicely, now it's time to refactor the code :)

resources/views/cruds/datagrids/filters/_array.blade.php
```diff
@@ -11,15 +11,29 @@
...
-            ],
-            $filterService->single($field['field'] . '_option'), [
-                'class' => 'form-control entity-list-filter',
-        ]) !!}
+        @if (isset($field['withChildren']) && $field['withChildren'] === true )
+            {!! Form::select(
+                $field['field'] . '_option',
+                [
+                    '' => __('crud.filters.options.include'),
+                    'children' => __('crud.filters.options.children'),
```
ilestis, line 19: Lots of copy-paste here when the only difference is if children is included in the list ;-)

ilestis (changes requested):
One last change to remove the last copy-pasting :)

ilestis (approved):

PR #481, opened 2022-08-12
Migrated the campaing roles table to datagrid2

spitfire305:


app/Http/Controllers/CampaignRoleController.php
```diff
@@ -151,11 +168,13 @@ public function update(StoreCampaignRole $request, CampaignRole $campaignRole)
      * @param  \App\Models\CampaignRole  $campaignRole
      * @return \Illuminate\Http\Response
      */
-    public function destroy(CampaignRole $campaignRole)
+    public function destroy(CampaignRole $campaignRole, bool $bulk = false)
```
ilestis, line 171: Feels "wrong" to have a controller function being called by another controller function. Since all we're doing it $campaigRole->delete(), we can call that directly in the bulk process instead of complicating the controller's destroy function.

app/Http/Controllers/CampaignRoleController.php
```diff
@@ -261,4 +280,39 @@ public function toggle(CampaignRole $campaignRole, int $entityType, int $action)
...
+        $count = 0;
+        foreach ($models as $id) {
+            /** @var CampaignRole $plugin */
+            $plugin = CampaignRole::find($id);
+            if (empty($plugin)) {
+                continue;
+            }
+
+            if ($action === 'delete' && !$plugin->isAdmin() && !$plugin->isPublic()) {
+                $this->destroy($plugin, true);
```
ilestis, line 309: Replace this by $plugin->delete() to simplify the code

app/Http/Controllers/CampaignRoleController.php
```diff
@@ -261,4 +280,39 @@ public function toggle(CampaignRole $campaignRole, int $entityType, int $action)
...
+
+    /**
+     * @return \Illuminate\Http\RedirectResponse
+     * @throws \Psr\Container\ContainerExceptionInterface
+     * @throws \Psr\Container\NotFoundExceptionInterface
+     */
+    public function bulk()
+    {
+        $campaign = CampaignLocalization::getCampaign();
+        $this->authorize('recover', $campaign);
```
ilestis, line 292: The "roles" permission should be used, not "recover" ;-)

app/Renderers/Layouts/Campaign/CampaignRole.php
```diff
@@ -0,0 +1,110 @@
...
+                        $html = '<a href="' . route('campaign_roles.show', ['campaign_role' => $model])
+                            . '" title="'
+                            . __('campaigns.roles.actions.permissions')
+                            . '">'
+                            . $model->permissions->whereNull('entity_id')->count()
+                            . '</a>';
+                    }
+                    if ($model->is_public && !$campaign->is_public && $model->permissions->whereNull('entity_id')->count() > 0) {
+
+                        $html = $html
```
ilestis, line 62: You can replace `$html = $html .` with simply `$html .= '<div>...'`

resources/views/campaigns/roles/_table.blade.php
```diff
@@ -0,0 +1,44 @@
+@if (empty($rows) && isset($empty))
```
ilestis, line 1: This blade file shouldn't be needed as we have `views/layouts/datagrids/table.blade.php` now :)

ilestis (changes requested):
Small tweaks

app/Http/Controllers/CampaignRoleController.php
```diff
@@ -306,7 +304,7 @@ public function bulk()
             }
 
             if ($action === 'delete' && !$plugin->isAdmin() && !$plugin->isPublic()) {
-                $this->destroy($plugin, true);
+                $plugin->delete();
```
ilestis, line 307: Just noticed that the variable is still named `$plugin` instead of what it is :)

ilestis (changes requested):
Last change, renaming the $plugin variable into $role ;-)

ilestis (approved):

PR #482, opened 2022-08-17
Timeline eras manual order

spitfire305:


app/Models/Timeline.php
```diff
@@ -182,6 +182,11 @@ public function menuItems(array $items = []): array
             'route' => 'timelines.timelines',
             'count' => $this->descendants()->count()
         ];
+        $items['second']['reorder'] = [
```
ilestis, line 185: Missing a permission check ;-)

app/Services/TimelineService.php
```diff
@@ -57,4 +58,31 @@ public function reorderEra(TimelineEra $era, array $ids)
...
+    {
+        $ids = $request->get('timeline_era');
+        if (empty($ids)) {
+            return false;
+        }
+
+        $position = 1;
+        foreach ($ids as $id) {
+            /** @var TimelineEra $link */
+            $link = TimelineEra::where('id', $id)->first();
```
ilestis, line 76: You can replace `where('id', $id)->first()` with simply `find($id)`. Also renaming `$link` to the correct model will make reading the code easier to understand :)

database/migrations/2022_08_15_193748_update_timeline_eras_add_position.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('timeline_eras', function (Blueprint $table) {
+            $table->unsignedTinyInteger('position');
```
ilestis, line 17: If you try and create a new era on a timeline, do you get an error? I have a feeling this is missing ->nullable() ;-)

resources/views/timelines/eras/reorder.blade.php
```diff
@@ -0,0 +1,70 @@
...
+                        {!! Form::hidden('timeline_era[]', $era->id) !!}
+                        <div class="dragger">
+                            <span class="fa-solid fa-ellipsis-v visible-md visible-lg"></span>
+                            <div class="visible-xs visible-sm">
+                                <span class="fa-solid fa-arrow-up"></span><br />
+                                <span class="fa-solid fa-arrow-down"></span>
+                            </div>
+                        </div>
+                        <div class="name">
+                            {!! $era->name !!}
```
ilestis, line 43: Wasn't part of the original ticket, but I think it's worth adding 
```html
 <span class="text-sm">
                {!! $era->ages()!!}
            </span>
```
After the name, like on the overview page

ilestis (changes requested):
Small changes, good job!

app/Models/Timeline.php
```diff
@@ -182,11 +182,13 @@ public function menuItems(array $items = []): array
             'route' => 'timelines.timelines',
             'count' => $this->descendants()->count()
         ];
-        $items['second']['reorder'] = [
-            'name' => 'timelines.fields.reorder',
-            'route' => 'timeline-eras.reorder',
-            'count' => $this->descendants()->count()
-        ];
+        if (auth()->check() && auth()->user()->can('permission', $this)) {
```
ilestis, line 185: `permission` is the wrong check for this page ;-)

ilestis (changes requested):
So close! Wrong permission check :)

ilestis (approved):

PR #483, opened 2022-08-17
Mentions now show the new options when the search results dont fully …

spitfire305:
…match the query.

### DIFF
--- a/app/Services/SearchService.php
+++ b/app/Services/SearchService.php
@@ -256,22 +256,19 @@ class SearchService
                 'advanced_mention' => Mentions::advancedMentionHelper($model->name),
             ];
         }
-
-        if (empty($searchResults) && $this->new) {
+        if (!$this->new) {
+            return $searchResults;
+        } elseif (empty($searchResults) && $this->new) {
             return $this->newOptions();
         }
 
         foreach ($searchResults as $result) {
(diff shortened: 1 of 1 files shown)


app/Services/SearchService.php
```diff
@@ -261,7 +261,13 @@ public function find()
...
         }
 
-        return $searchResults;
+        foreach ($searchResults as $result) {
+            if (strtolower($result['name']) == $cleanTerm) {
+                return $searchResults;
+            }
+        }
+
+        return $searchResults + $this->newOptions();
```
ilestis, line 270: This now offers the "new" option in the entity filters for example (add an entity to an ability), which isn't exactly what we want, and shows only one result if I type in "goblin" when I have a race called "goblin" and "hobgoblin", which also isn't great :)

ilestis (changes requested):
Needs some different logic

app/Services/SearchService.php
```diff
@@ -261,6 +261,16 @@ public function find()
             return $this->newOptions();
         }
 
+        foreach ($searchResults as $result) {
```
ilestis, line 264: We can save resources by only doing an if first that checks that it's not New, to return the results directly, rather than a foreach

ilestis (changes requested):

ilestis (approved):
Awesome!

PR #485, opened 2022-08-24
Users can now define multiple locations for a race

spitfire305:


app/Models/Concerns/HasFilters.php
```diff
@@ -122,6 +123,8 @@ public function scopeFilter(Builder $query, array $params = []): Builder
                     $this->filterRaces($query, $value);
                 } elseif ($key == 'location_id') {
                     $this->filterLocations($query, $value, $key);
+                } elseif ($key == 'locations_id') {
```
ilestis, line 126: Confusing that the variable is singular and the function is plural, and vice versa :) would be cleaner to have a check `if ($this instanceof Race) { filterRaceLocations() }`

app/Observers/RaceObserver.php
```diff
@@ -4,9 +4,85 @@
...
+            $existing[$location->id] = $location->id;
+            $unique[$location->id] = $location->id;
+        }
+
+        if (!empty($recreate)) {
+            $race->locations()->attach($recreate);
+        }
+
+        $locations = request()->get('locations', []);
+        $newFamilies = [];
```
ilestis, line 50: $newFamilies? :D

app/Services/Entity/ConnectionService.php
```diff
@@ -253,6 +254,16 @@ protected function loadOrganisations(): self
         }
         return $this;
     }
+    protected function loadRaces(): self
```
ilestis, line 257: Missing empty line between previous and new function

resources/views/cruds/fields/locations.blade.php
```diff
@@ -0,0 +1,11 @@
+@if (!$campaignService->enabled('families'))
```
ilestis, line 1: Wrong module check I think

resources/views/entities/components/profile/races.blade.php
```diff
@@ -10,6 +10,19 @@
         <i class="fa-solid fa-chevron-down"></i>
         {{ __('crud.tabs.profile') }}
     </div>
+    @if (!$model->locations->isEmpty())
+            <div class="element profile-location">
+                <div class="title">{{ __('races.fields.locations') }}</div>
```
ilestis, line 15: Variable name needs fixing

app/Models/Race.php
```diff
@@ -207,6 +212,27 @@ public function filterableColumns(): array
...
+     * @return bool
+     */
+    public function showProfileInfo(): bool
+    {
+
+        if ($this->locations) {
+            return true;
+        }
+
+        return false;
```
ilestis, line 236: The check on the type has gone. Might be simpler to return parent::showProfile() instead of return false, to use the basic showProfile function from the MiscModel

ilestis (changes requested):
Nice job! A tiny few changes :)

resources/views/entities/components/profile/races.blade.php
spitfire305: What variable?

ilestis: Woops, this was for another area in the code, but looks fixed 🙈

PR #486, opened 2022-08-26
Intercalary months now appear as such when displaying dates.

spitfire305:


app/Traits/CalendarDateTrait.php
```diff
@@ -51,7 +51,22 @@ public function hasCalendarButNoAccess(): bool
...
+            }
+            if ($monthCount == $reminder->month) {
+
+                if ($month["type"] == "intercalary") {
+                    return $reminder->year . '-' . $month["name"] . '-' . $reminder->day;
+                }
+                return $reminder->year . '-' . $count . '-' . $reminder->day;
+            }
+            $monthCount++;
+        }
```
ilestis, line 69: In case the calendar is broken and has no months, it's still worth having a fallback return at the end of the function what uses the old syntax :)

ilestis (changes requested):
Nice job! Just missing a fallback for invalid calendars that, in theory, shouldn't be possible, but reality is cruel and unforgiving

ilestis (approved):

PR #487, opened 2022-08-31
Posts move to another entity

spitfire305:


app/Http/Controllers/Entity/PostMoveController.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @param TransformEntityRequest $request
+     * @param Entity $entity
+     * @return \Illuminate\Http\RedirectResponse
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function move(MovePostRequest $request, Entity $entity, EntityNote $entityNote)
+    {
+
```
ilestis, line 60: Missing permission checks, an unlogged user or one without the correct permission could call this page right now :-)

app/Http/Controllers/Entity/PostMoveController.php
```diff
@@ -0,0 +1,76 @@
...
+     * @param Entity $entity
+     * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function index(Entity $entity, EntityNote $entityNote)
+    {
+        $this->authorize('view', $entity->child);
+        $campaign = CampaignLocalization::getCampaign()->entities()->get();
+
+        return view('entities.pages.move.entity_notes.index', compact(
```
ilestis, line 45: We already have a folder for entity-notes, might as well put the `move` page there instead of a new folder

ilestis (changes requested):
Some tweaks required

app/Http/Controllers/Entity/PostMoveController.php
```diff
@@ -59,12 +53,20 @@ public function move(MovePostRequest $request, Entity $entity, EntityNote $entit
...
 
         $newEntity = Entity::where(['id' => $request['entity']])->first();
+        $this->authorize('update', $newEntity->child);
         try {
             $newNote = $this->service
                 ->moveEntityNote($entityNote, $request->only('entity', 'copy'));
+            if (isset($request['copy'])) {
+                return redirect()
+                ->route($newEntity->pluralType() . '.show', [$newEntity->child->id, '#post-' . $newNote->id])
+                ->with('success', __('entities/notes.move.copy_success', ['name' => $newNote->name,
```
ilestis, line 63: Lots of copy-paste here when all we need is to change the success messsage. This could be refactored to be
```
$success = 'move_success';
if ($request->has('copy')) {
    $success = 'copy_success';
}
return redirect()...
```

ilestis (changes requested):
Small code refactoring, but also one thing that is missing: when copying instead of moving, I don't think it currently copies the permissions of the entity note :)

ilestis (approved):
Awesome!

PR #488, opened 2022-08-31
Campaigns shadowbanned

spitfire305:


app/Http/Controllers/CampaignController.php
```diff
@@ -222,6 +222,16 @@ public function leave()
         }
     }
 
+    /**
```
ilestis, line 225: Seems like this should be removed? :)

app/Models/Campaign.php
```diff
@@ -98,6 +99,7 @@ class Campaign extends MiscModel
         'ui_settings',
         'settings',
         'is_open',
+        'is_pinned',
```
ilestis, line 102: Is pinned? 🤔

app/Services/CampaignService.php
```diff
@@ -128,6 +128,39 @@ public function leave(Campaign $campaign)
...
+                $campaign,
+                'hidden',
+                'eye-slash',
+                'yellow',
+                [
+                    'campaign' => $campaign->name,
+                    'link' => $campaign->getMiddlewareLink()
+                ]
+            );
+        } elseif (!$campaign->isHidden()) {
```
ilestis, line 150: else should be enough, since it's either true or false in the if statement

resources/views/front/campaigns.blade.php
```diff
@@ -72,9 +72,11 @@
                 @else
                 <div class="row">
                     @foreach ($campaigns as $camp)
-                        <div class="col-sm-6">
-                            @include('front._campaign', ['campaign' => $camp, 'featured' => false])
-                        </div>
+                        @if (!$camp->isHidden())
```
ilestis, line 75: This check should be done in the database query, not in the view. Otherwise we'll have pages with less than 15 campaigns visible, and have a wrong pagination count.

resources/views/users/profile.blade.php
```diff
@@ -103,9 +103,11 @@ class="profile-subscriber" title="Kobold" />
 
                             <div class="row">
                                 @foreach ($campaigns as $campaign)
-                                    <div class="col-xl-6 col-12">
-                                        @include('front._campaign', ['campaign' => $campaign, 'featured' => false])
-                                    </div>
+                                    @if (!$campaign->isHidden())
```
ilestis, line 106: Same thing here, the check should be done on the query, not in the UI.

ilestis (changes requested):
Move some of the logic to the database and out of the UI.

app/Http/Controllers/CampaignController.php
spitfire305: If I remove it, the notifications stop working, the job gets processed but the notifications are never sent to the campaign admins.

app/Http/Controllers/FrontController.php
```diff
@@ -164,8 +164,9 @@ public function partners()
      */
     public function campaigns(Request $request)
     {
-        $featured = Campaign::public()->front()->featured()->get();
+        $featured = Campaign::public()->where('is_hidden', 0)->front()->featured()->get();
```
ilestis, line 167: It would be even better to just add the condition to the `frontScope` on campaigns, so we don't have to remember to add it in multiple places ;-)

app/Http/Controllers/User/ProfileController.php
```diff
@@ -12,7 +12,7 @@ class ProfileController extends Controller
      */
     public function show(User $user)
     {
-        $campaigns = $user->campaigns()->public()->front()->paginate();
+        $campaigns = $user->campaigns()->public()->where('is_hidden', 0)->front()->paginate();
```
ilestis, line 15: Same as above

ilestis (changes requested):
One last fix: put the query in the scope, to avoid duplicating code

ilestis (approved):

PR #491, opened 2022-09-08
Changes to subscription cancellation form, now asks users for a custo…

spitfire305:
…m cancellation reason, also changed the email from which the notifications are sent.

### DIFF
--- a/app/Mail/Subscription/User/CancelledUserSubscriptionMail.php
+++ b/app/Mail/Subscription/User/CancelledUserSubscriptionMail.php
@@ -41,7 +41,7 @@ class CancelledUserSubscriptionMail extends Mailable
     public function build()
     {
         return $this
-            ->from(['address' => 'hello@kanka.io', 'name' => 'Kanka Team'])
+            ->from(['address' => 'no-reply@kanka.io', 'name' => 'Kanka Team'])
             ->subject('Subscription: Cancelled ' . $this->user->patreon_pledge)
             ->view('emails.subscriptions.cancelled.user-html');
     }
--- a/app/Mail/Subscription/User/CancelledUserSubscriptionMail.php
+++ b/app/Mail/Subscription/User/CancelledUserSubscriptionMail.php
@@ -41,7 +41,7 @@ class CancelledUserSubscriptionMail extends Mailable
     public function build()
     {
         return $this
-            ->from(['address' => 'no-reply@kanka.io', 'name' => 'Kanka Team'])
+            ->from(['address' => 'hello@kanka.io', 'name' => 'Kanka Team'])
             ->subject('Subscription: Cancelled ' . $this->user->patreon_pledge)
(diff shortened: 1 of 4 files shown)


app/Mail/Subscription/User/CancelledUserSubscriptionMail.php
```diff
@@ -41,7 +41,7 @@ public function __construct(User $user, string $reason = null)
     public function build()
     {
         return $this
-            ->from(['address' => 'hello@kanka.io', 'name' => 'Kanka Team'])
+            ->from(['address' => 'no-reply@kanka.io', 'name' => 'Kanka Team'])
```
ilestis, line 44: This is the reverse, no-reply doesn't exist and gets flagged as spam. I remembered wrong, it's other emails that are still being sent as no-reply instead of hello. You can revert this change and ignore the other places, I'll create a new ticket.

resources/assets/js/subscription.js
```diff
@@ -26,18 +26,6 @@ function initStripe() {
     elements = stripe.elements();
 }
 
-function initCancelListener()
```
ilestis, line 29: Great but `initCancelListener()` is still being called from another part of this script :)

resources/views/settings/subscription/change.blade.php
```diff
@@ -208,28 +209,31 @@
...
-], null, ['class' => 'form-control mb-5 select-reveal-field', 'data-change-target' => '#cancel-reason-custom']) !!}
+], null, ['class' => 'form-control mb-5']) !!}
             {!! Form::textarea(
                 'reason_custom',
                 null,
                 [
                     'placeholder' => __('settings.subscription.placeholders.reason'),
                     'class' => 'form-control',
-                    'style' => 'display: none',
+                    'style' => 'display: inline',
```
ilestis, line 236: This whole style line can be removed

resources/views/settings/subscription/change.blade.php
```diff
@@ -208,28 +209,31 @@
             </p>
     @else
         {!! Form::open(['route' => ['settings.subscription.subscribe'], 'method' => 'POST', 'id' => 'cancellation-confirm', 'class' => 'subscription-form']) !!}
-
+        @php
+            $endDate = date($user->date_format, $user->upcomingInvoice()->period_end);
```
ilestis, line 213: I don't believe that `date()` uses the user's date format. Check how it's done for reminders, by searching for `dateRenderer` :)

ilestis (changes requested):
Small tweaks

resources/views/settings/subscription/change.blade.php
spitfire305: It does! I tried changing my date format and it works as intended for all the formats.

ilestis (approved):

PR #498, opened 2022-09-21
Added the ability for users to bulk unmirror relations.

spitfire305:


app/Services/BulkService.php
```diff
@@ -415,6 +415,14 @@ protected function updateRelations(array $filledFields)
                 $this->count++;
                 $this->total++;
             }
+            if (request()->unmirror && $relation->mirror) {
```
ilestis, line 418: Using `request()` means we can't ever *unit test* this class (since request() is only available when executing the whole framework and not when testing just a small feature).

Instead, in `editing()`, just before calling `$this->updateRelations()`, check the `$fields` array for the options to build an array as a second parameter for the `updateRelations()` call.

ilestis (changes requested):
We can't unit test `request()` so we need to remove that from the function call and find a workaround :)

app/Services/BulkService.php
```diff
@@ -259,7 +259,11 @@ public function editing(array $fields, Bulk $bulk): int
         $tagIds = Arr::get($fields, 'tags', []);
 
         if ($this->entityName === 'relations') {
-            return $this->updateRelations($filledFields);
+            $mirrorOptions = [];
+            $mirrorOptions['unmirror'] = request()->unmirror;
```
ilestis, line 263: request() still being called in BulkService, so not testable :P `$fields` equals `request()->all()` so we can test on `$mirroOptions['unmirror'] = (bool) Arr::get($fields, 'unmirror', '0');`

ilestis (changes requested):
Request() still being called in BulkService :D

ilestis (approved):
Perfect, thanks!

PR #502, opened 2022-09-26
CSS Themes now load the fonts section of the plugins before the rest …

spitfire305:
…of the CSS code.

### DIFF
--- a/app/Services/Caches/CampaignCacheService.php
+++ b/app/Services/Caches/CampaignCacheService.php
@@ -245,6 +245,9 @@ class CampaignCacheService extends BaseCache
         /** @var CampaignPlugin $plugin */
         foreach ($plugins as $plugin) {
             $theme .= "/** plugin: " . e($plugin->name) . " #" . e($plugin->version->version) . " **/\n";
+            $theme .= $plugin->version->fonts . "\n\n";
+        }
+        foreach ($plugins as $plugin) {
             $theme .= $plugin->version->content . "\n\n";
         }
 


app/Services/Caches/CampaignCacheService.php
```diff
@@ -245,6 +245,9 @@ public function themes(): string|bool
         /** @var CampaignPlugin $plugin */
         foreach ($plugins as $plugin) {
             $theme .= "/** plugin: " . e($plugin->name) . " #" . e($plugin->version->version) . " **/\n";
```
ilestis, line 247: Would be great to only have a comment in the css if the fonts field isn't empty. At the moment it's adding 16 useless lines on my local config :P

app/Services/Caches/CampaignCacheService.php
```diff
@@ -245,6 +245,9 @@ public function themes(): string|bool
...
+            $theme .= $plugin->version->fonts . "\n\n";
+        }
+        foreach ($plugins as $plugin) {
             $theme .= $plugin->version->content . "\n\n";
```
ilestis, line 251: Would be great to have a comment in the css file again to quickly find declarations for debugging

ilestis (changes requested):
Small change + missing `fonts` declaration (see the checks tab)

app/Services/Caches/CampaignCacheService.php
```diff
@@ -244,10 +244,13 @@ public function themes(): string|bool
             ->get();
         /** @var CampaignPlugin $plugin */
         foreach ($plugins as $plugin) {
-            $theme .= "/** plugin: " . e($plugin->name) . " #" . e($plugin->version->version) . " **/\n";
-            $theme .= $plugin->version->fonts . "\n\n";
+            $theme .= "/** plugin: " . e($plugin->name) . " #" . e($plugin->version->version) . " fonts **/\n";
```
ilestis, line 247: Sorry for not making it clear, only add the code and comment **if** there is a font. Otherwise my local setup starts with 16 comments but no content, which isn't useful :)

ilestis (changes requested):
Was unclear about what I wished for

ilestis (approved):

PR #503, opened 2022-09-27
Saving and copying an entity now also copies the image.

spitfire305:


resources/views/cruds/fields/image.blade.php
```diff
@@ -21,7 +21,7 @@
             {!! Form::file('image', array('class' => 'image form-control')) !!}
         </div>
         <div class="form-group">
-            {!! Form::text('image_url', null, ['placeholder' => __('crud.placeholders.image_url'), 'class' => 'form-control']) !!}
+            {!! Form::text('image_url', (!empty($source) ? $source->getImageUrl() : ''), ['placeholder' => __('crud.placeholders.image_url'), 'class' => 'form-control']) !!}
```
ilestis, line 24: This works, but it also adds the "default image" if the source has no image and there is one setup in the campaign (World > Default thumbnails). Also on prod it links to the "images.kanka.io" thumbnail server instead of the full image source :) `getOriginalImageUrl()` is the method you want ;-)

Lastly, since we're not using PHP 8, we can simplify the code to simply `$source?->getOriginalImageUrl()` instead of doing a ternary if

ilestis (changes requested):
Very close :)

resources/views/cruds/fields/image.blade.php
```diff
@@ -21,7 +21,7 @@
             {!! Form::file('image', array('class' => 'image form-control')) !!}
         </div>
         <div class="form-group">
-            {!! Form::text('image_url', (!empty($source) ? $source->getImageUrl() : ''), ['placeholder' => __('crud.placeholders.image_url'), 'class' => 'form-control']) !!}
+            {!! Form::text('image_url', $source?->getOriginalImageUrl(), ['placeholder' => __('crud.placeholders.image_url'), 'class' => 'form-control']) !!}
```
ilestis, line 24: I lied, this doesn't work and generates an error 🙈  need to test that the source has an image first before calling the function

ilestis (changes requested):
I was wrong, this crashes on copying an entity with no image

ilestis (approved):
Awesome!

PR #504, opened 2022-09-28
Corrected a bug where items werent showing the tree view when nested …

spitfire305:
…views are enabled by default.


ilestis (approved):
Easy fix, nice!

PR #505, opened 2022-09-30
Added a confirmation check for users that want to delete a campaign

spitfire305:


app/Http/Requests/DeleteCampaign.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        $user = Auth::user();
```
ilestis, line 28: Not sure why we need the $user? :P

resources/views/campaigns/_menu.blade.php
```diff
@@ -271,8 +271,11 @@
 
                         @if (auth()->user()->can('delete', $campaign))
                             <p class="mt-5">{!! __('campaigns.destroy.confirm', ['campaign' => '<strong>' . $campaign->name . '</strong>']) !!}
-
+                            <p class="help-block"> {{ __('campaigns.destroy.hint', ['code' => 'delete']) }} </p>
```
ilestis, line 274: Place a `<code>` tag around the delete code to make it clearer for users

resources/views/campaigns/_menu.blade.php
```diff
@@ -271,8 +271,11 @@
...
                             {!! Form::open(['method' => 'DELETE', 'route' => ['campaigns.destroy', $campaign->id]]) !!}
+                            <div class="form-group required">
+                                {!! Form::text('delete',null, ['class' => 'form-control','required']) !!}
```
ilestis, line 277: Missing space after some of the `,`

ilestis (changes requested):
Tiny tweaks

ilestis (approved):

PR #508, opened 2022-10-04
Added the ability for users to filter quests based on their entities …

spitfire305:
…or on the roles from their entities

### DIFF
--- a/app/Http/Controllers/Search/MiscController.php
+++ b/app/Http/Controllers/Search/MiscController.php
@@ -353,15 +353,4 @@ class MiscController extends Controller
 
         return Response::json($formatted);
     }
-
-    /**
-     * @param Request $request
-     * @return mixed
-     */
-    public function questElements(Request $request)
-    {
-        $term = trim($request->q);
-        $exclude = $request->has('exclude') ? [$request->get('exclude')] : [];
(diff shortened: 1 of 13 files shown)


app/Http/Controllers/Search/MiscController.php
```diff
@@ -353,4 +353,15 @@ protected function buildSearchResults($term, $class, array $excludes = [])
 
         return Response::json($formatted);
     }
+
+    /**
+     * @param Request $request
+     * @return mixed
+     */
+    public function questElements(Request $request)
```
ilestis, line 361: Not needed, we already have this with filtering on some entity types which can't be in quests

ilestis (changes requested):
Small changes to re-use existing code rather than duplicate existing features :) nice job

app/Observers/QuestElementObserver.php
```diff
@@ -34,5 +36,6 @@ public function saving(QuestElement $questElement)
     public function creating(QuestElement $questElement)
     {
         $questElement->created_by = auth()->user()->id;
+        QuestCache::clearSuggestion();
```
ilestis, line 39: You can replace both of these calls by calling it only once in `saved(...)` :)

ilestis (changes requested):
Tiny last change :)

ilestis (approved):

PR #509, opened 2022-10-04
Attributes endless loop bug

spitfire305:


ilestis (approved):
Looks good, well done :)

PR #513, opened 2022-10-12
Attribute values now display correctly when mentioned inside an entit…

spitfire305:
…y mention.

### DIFF
--- a/app/Services/MentionsService.php
+++ b/app/Services/MentionsService.php
@@ -121,11 +121,10 @@ class MentionsService
             return Attributes::parse($attribute);
         }
 
-        $this->text = (string) $attribute->value;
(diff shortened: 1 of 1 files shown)


app/Services/MentionsService.php
```diff
@@ -122,6 +122,7 @@ public function mapAttribute(Attribute $attribute)
         }
 
         $this->text = (string) $attribute->value;
+        $this->text = Attributes::parse($attribute);
```
ilestis, line 125: You're setting $this->text twice, and calling `Attributes::parse()` twice, which seems very redundant (and bad for performance). Can you clean up the code to only call these things once? And also add a comment in the code as to why we are calling the methods in this order, and not the old order :)

ilestis (approved):

PR #515, opened 2022-10-15
Refactored the marker groups forms and implemented updated datagrid 2…

spitfire305:
… and drag and drop control features.


app/Http/Controllers/Maps/MapGroupController.php
```diff
@@ -5,12 +5,41 @@
...
+use App\Http\Requests\ReorderGroups;
+use App\Datagrids\Actions\GroupDatagridActions;
 use App\Models\Campaign;
 use App\Models\Map;
 use App\Models\MapGroup;
 
 class MapGroupController extends Controller
 {
+    /** @var string|null The datagrid controlling the bulk actions*/
+    protected $datagridActions = GroupDatagridActions::class;
```
ilestis, line 18: Never used?

app/Datagrids/Actions/GroupDatagridActions.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace App\Datagrids\Actions;
```
ilestis, line 3: Unclear why this file is needed, it's used nowhere

app/Http/Controllers/Maps/MapGroupController.php
```diff
@@ -5,12 +5,41 @@
...
+
+    /**
+     * Index
+     */
+    public function groups(Map $map)
```
ilestis, line 23: Please rename to index() :)

app/Http/Controllers/Maps/MapGroupController.php
```diff
@@ -5,12 +5,41 @@
...
+    {
+        $campaign = CampaignLocalization::getCampaign();
```
ilestis, line 25: I can call this page in an incognito window. Only a user who can edit the map should have access to it.

app/Http/Controllers/Maps/MapGroupController.php
```diff
@@ -5,12 +5,41 @@
...
+        $options = ['map' => $map];
+        $model = $map;
+
+        Datagrid::layout(\App\Renderers\Layouts\Map\Group::class)
+            ->route('maps.groups', $options);
+        $rows = $map
```
ilestis, line 31: This can be improved :) 
image

resources/views/maps/groups/edit.blade.php
```diff
@@ -27,20 +27,47 @@
         @endif
         <div class="panel-body">
```
ilestis, line 28: Can't use "panel-body" when in a modal, it creates a weird border. Need to use "modal-body"
image

resources/views/maps/groups/create.blade.php
```diff
@@ -27,20 +27,47 @@
         @endif
         <div class="panel-body">
```
ilestis, line 28: Can't use `panel-body` in a modal, needs to be `modal-body` when it's an ajax request, and panel-body when not. 
image

resources/views/maps/groups/create.blade.php
```diff
@@ -27,20 +27,47 @@
...
-            <div class="form-group">
-                <div class="submit-group">
-                    <button class="btn btn-success">{{ __('crud.save') }}</button>
-                    @includeWhen(!request()->ajax(), 'partials.or_cancel')
-                </div>
-                <div class="submit-animation" style="display: none;">
-                    <button class="btn btn-success" disabled><i class="fa-solid fa-spinner fa-spin"></i></button>
+            {!! Form::open(['route' => ['maps.map_groups.store', $map], 'method' => 'POST', 'data-shortcut' => 1]) !!}
+                @include('maps.groups._form', ['model' => null, 'map' => $map])
+                <div class="pull-right">
```
ilestis, line 32: This div and the next one can be merged into a single div with two css classes

resources/views/maps/groups/edit.blade.php
```diff
@@ -27,20 +27,47 @@
...
-
-            <div class="form-group">
-                <div class="submit-group">
-                    <button class="btn btn-success">{{ __('crud.save') }}</button>
-                    @includeWhen(!request()->ajax(), 'partials.or_cancel')
-                </div>
-                <div class="submit-animation" style="display: none;">
-                    <button class="btn btn-success" disabled><i class="fa-solid fa-spinner fa-spin"></i></button>
+                @include('maps.groups._form')
+                <div class="pull-right">
```
ilestis, line 32: Merge the two divs together

app/Models/Map.php
```diff
@@ -305,6 +310,20 @@ public function groupOptions(): array
         return $options;
     }
 
+    /**
+     * @return array|string[]
+     */
+    public function groupPositionOptions(): array
+    {
+        $options = [null => ''];
```
ilestis, line 318: Overwritten by the next line

app/Models/MapGroup.php
```diff
@@ -25,7 +26,17 @@
  */
 class MapGroup extends Model
 {
-    use VisibilityIDTrait, Blameable, Paginatable;
+    use VisibilityIDTrait, Blameable, Paginatable, SortableTrait;
+
+    protected $sortable = [
```
ilestis, line 31: I think only one of these is needed. Either $sortable or $sortableColumns, but not both :)

resources/views/maps/groups/create.blade.php
ilestis, line 32: Also missing a "cancel" button that either closes the modal, or goes back when the url is opened in a new tab. Same as we have on other modals.
image

resources/views/maps/groups/_form.blade.php
```diff
@@ -16,5 +16,5 @@
 
 <div class="form-group">
     <label>{{ __('maps/groups.fields.position') }}</label>
-    {!! Form::number('position', null, ['placeholder' => __('maps/groups.placeholders.position'), 'class' => 'form-control', 'maxlength' => 3]) !!}
+    {!! Form::select('position', isset($map) ? $map->groupPositionOptions() : [], (!empty($group) ? $group : null), ['class' => 'form-control']) !!}
```
ilestis, line 19: Default when adding a new group should be the last option being selected, rather than the first.

ilestis: Also, the "name" field is missing the `required` attribute, to avoid sending the form without filling it out.

app/Renderers/Layouts/Map/Group.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Available columns
+     * @return array[]
+     */
+    public function columns(): array
+    {
+        $columns = [
+            'name' => [
+                'key' => 'name',
+                'label' => 'crud.fields.name',
```
ilestis, line 19: The "name" can be a link to the `edit` form.

app/Models/Map.php
```diff
@@ -259,6 +259,11 @@ public function menuItems(array $items = []): array
             'route' => 'maps.maps',
             'count' => $this->maps()->count()
         ];
+        $items['third']['groups'] = [
```
ilestis, line 262: Place the link in the second box instead of the "third" box.

app/Models/Map.php
```diff
@@ -259,6 +259,11 @@ public function menuItems(array $items = []): array
...
+            'name' => 'maps.panels.groups',
+            'route' => 'maps.groups',
+            'count' => $this->groups()->count()
```
ilestis, line 265: Missing the permission check

routes/campaign.php
```diff
@@ -32,6 +32,9 @@
 Route::post('/maps/{map}/{map_marker}/move', 'Maps\MapMarkerController@move')->name('maps.markers.move');
 Route::get('/maps/tree', 'Maps\MapController@tree')->name('maps.tree');
 //Route::get('/maps/{map}/map-points', 'Maps\MapController@mapPoints')->name('maps.map-points');
+Route::get('/maps/{map}/groups', 'Maps\MapGroupController@groups')->name('maps.groups');
```
ilestis, line 35: Unnecessary if we're using the existing index route

ilestis (changes requested):
Nice job! I have a few changes needed to make it perfect.

lang/en/maps/groups.php
```diff
@@ -33,8 +33,9 @@
         'until' => 'Create up to :max groups to each map.',
     ],
     'placeholders'  => [
-        'name'      => 'Shops, Treasure, NPCs',
-        'position'  => 'Last',
+        'name'           => 'Shops, Treasure, NPCs',
+        'position'       => 'First',
+        'position_list'  => 'After ',
```
ilestis, line 38: Last tweak, have :name as a param, because in some language the name will be before the translation for "after" :)

ilestis (changes requested):
Tiny last change for our translators

ilestis (approved):
🥳

PR #516, opened 2022-10-18
Moved layers from the maps form to a new layers subpage of maps in da…

spitfire305:
…tagrid2


app/Http/Controllers/Maps/MapLayerController.php
```diff
@@ -23,6 +29,31 @@ public function index(Map $map)
         return redirect()->route('maps.show', $map);
     }
 
+    /**
+     * Index
+     */
+    public function layers(Map $map)
```
ilestis, line 35: You can remove the old `index()` function and replace it with this

app/Http/Controllers/Maps/MapLayerController.php
```diff
@@ -135,7 +195,79 @@ public function destroy(Map $map, MapLayer $mapLayer)
...
+            ->with('success', trans_choice('maps/layers.bulks.' . $action, $count, ['count' => $count]))
+        ;
+    }
+
+    /**
+     * Controls drag and drop reordering of map layers
+     */
+    public function reorder(ReorderLayers $request)
+    {
+        $order = 1;
```
ilestis, line 256: Missing a authorization check :)

app/Http/Controllers/Maps/MapLayerController.php
```diff
@@ -135,7 +195,79 @@ public function destroy(Map $map, MapLayer $mapLayer)
...
+        $ids = $request->get('layer');
+        foreach ($ids as $id) {
+            $layer = MapLayer::find($id);
+            if (empty($layer)) {
```
ilestis, line 260: If I edit the ids in the html before sending, I can re-order layers from another map 🙈  needs to be fixed too on groups

resources/views/maps/layers/index.blade.php
```diff
@@ -0,0 +1,37 @@
...
+            'model' => $model,
+            'breadcrumb' => [
+                ['url' => Breadcrumb::index('maps'), 'label' => __('maps.index.title')],
+                null
+            ]
+        ])
+        @include('maps._menu', ['active' => 'layers'])
+        <div class="entity-main-block">
+            @include('maps.panels.layers')
+            @includeWhen(!$rows->isEmpty(), 'maps.layers._reorder')
```
ilestis, line 34: For a better UX, would make more sense to only show the reorder block when there are at least 2 rows, rather than at least one. The same can be changed for groups.

routes/campaign.php
```diff
@@ -35,6 +35,10 @@
 Route::post('/maps/{map}/groups/bulk', 'Maps\MapGroupController@bulk')->name('maps.groups.bulk');
 Route::post('/maps/groups/reorder', 'Maps\MapGroupController@reorder')->name('maps.groups.reorder-save');
 
+Route::get('/maps/{map}/layers', 'Maps\MapLayerController@layers')->name('maps.layers');
```
ilestis, line 38: Not needed when re-using index()

routes/campaign.php
```diff
@@ -35,6 +35,10 @@
...
+Route::post('/maps/{map}/layers/bulk', 'Maps\MapLayerController@bulk')->name('maps.layers.bulk');
+Route::post('/maps/layers/reorder', 'Maps\MapLayerController@reorder')->name('maps.layers.reorder-save');
```
ilestis, line 40: I'm confused why `{map}` isn't included here. Should be, for permissions, as well as for the group reorder.

ilestis (changes requested):
Few tweaks and missing permission checks

app/Models/MapLayer.php
```diff
@@ -87,4 +93,52 @@ public function typeName(): string
         }
         return 'overlay_shown';
     }
+
+    /**
+     * Functions for the datagrid2
+     * @return string
+     */
+    public function deleteName(): string
```
ilestis, line 101: `displayName()` is not needed  when you simply return the name, see `resources/views/layouts/datagrid/actions.blade.php`

ilestis (changes requested):
Tiny tweak that I remembered

ilestis (approved):
Awesome!

PR #517, opened 2022-10-19
The Map Marker form now shows custom icon suggestions to the user bas…

spitfire305:
…ed on their other markers

### DIFF
--- a/app/Services/Caches/MapMarkerCacheService.php
+++ b/app/Services/Caches/MapMarkerCacheService.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace App\Services\Caches;
+
+use App\Models\MapMarker;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+
+class MapMarkerCacheService extends BaseCache
+{
+    /**
+     * @return array
+     */
+    public function iconSuggestion(): array
+    {
+        $key = $this->iconSuggestionKey();
+        if (Cache::has($key)) {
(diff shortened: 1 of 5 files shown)


app/Services/Caches/MapMarkerCacheService.php
```diff
@@ -0,0 +1,74 @@
...
+            'ra ra-skull',
+            'fa-solid fa-coins',
+            'ra ra-beer',
+            'fa-solid fa-map-marker-alt',
+            'fa-solid fa-thumbtack',
+            'ra ra-wooden-sign',
+            'fa-solid fa-map-pin'
+        ];
+
+        $data = MapMarker::select(DB::raw('custom_icon, MAX(created_at) as cmat'))
```
ilestis, line 34: This is building stats on the whole table, not on just the map markers of the campaign

app/Services/Caches/MapMarkerCacheService.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+
+    /**
+     * Type suggestion cache key
+     * @return string
+     */
+    protected function iconSuggestionKey(): string
+    {
+        return 'campaign_' . $this->campaign->id . '_map_map_marker_suggestions';
```
ilestis, line 72: Can skip the double `map_` in the key ;-)

ilestis (changes requested):
Missing a limitation to only search map markers of the current campaign

ilestis (approved):

PR #519, opened 2022-10-21
Refactored the marker creation tab into its own page with datagrid 2 …

spitfire305:
…support.


app/Models/MapMarker.php
```diff
@@ -141,6 +156,52 @@ public function isPolygon(): bool
...
+        $icon = '<i class="fa-solid fa-map-pin"></i>';
+
+        $campaign = CampaignLocalization::getCampaign();
+        if (!empty($this->custom_icon) && $campaign->boosted()) {
+            if (Str::startsWith($this->custom_icon, '<i ')) {
+                $icon = $this->custom_icon ;
+            } elseif (Str::startsWith($this->custom_icon, ['fa-', 'ra '])) {
+                $icon = ' <i class="' . $this->custom_icon . '" aria-hidden="true"></i>';
+            } elseif (Str::startsWith($this->custom_icon, '<?xml')) {
+                $icon = 'L.Util.template(<div class="custom-icon">' . $this->resizedCustomIcon() . '</div>)';
```
ilestis, line 194: This needs to be tested and fixed ;-)
image

resources/views/maps/form/_markers.blade.php
```diff
@@ -9,20 +9,23 @@
         <p>{{ __('maps.helpers.missing_image') }}</p>
     </div>
 @else
-    <p class="help-block">
-        {{ __('maps/markers.helpers.base') }}
-    </p>
+    @if (auth()->check() && !auth()->user()->settings()->get('tutorial_map_markers'))
+        <div class="alert alert-info tutorial">
+        <span>
+            <button type="button" class="close banner-notification-dismiss" data-dismiss="alert" aria-hidden="true" data-url="{{ route('settings.banner', ['code' => 'map_layers', 'type' => 'tutorial']) }}">×</button>
```
ilestis, line 15: Wrong code :-)

resources/views/maps/markers/create.blade.php
```diff
@@ -42,39 +42,9 @@
...
-                    </ul>
-                </div>
-            </div>
-            <div class="submit-animation" style="display: none;">
-                <button class="btn btn-success" disabled><i class="fa-solid fa-spinner fa-spin"></i></button>
+                @include('partials.footer_cancel', ['ajax' => null])
             </div>
+            @include('maps.markers._actions')
         </div>
     </div>
```
ilestis, line 49: Something is wrong in the HTML, it's creating this ugly line in the footer 
image
See how I fixed it in Layers

app/Models/MapMarker.php
```diff
@@ -141,6 +156,52 @@ public function isPolygon(): bool
...
+        return __('maps/markers.tabs.marker');
+    }
+
+    /**
+     * Determine the icon of the marker
+     * @return string
+     */
+    public function datagridMarkerIcon(): string
+    {
+        if ($this->icon == 5 || $this->icon == 4) {
```
ilestis, line 181: Showing a pin icon for circle and polygons
image

app/Models/MapMarker.php
```diff
@@ -141,6 +156,52 @@ public function isPolygon(): bool
...
+        } elseif ($this->isLabel()) {
+            return __('maps/markers.tabs.label');
+        } elseif ($this->isCircle()) {
+            return __('maps/markers.tabs.circle');
+        }
+        return __('maps/markers.tabs.marker');
+    }
+
+    /**
+     * Determine the icon of the marker
```
ilestis, line 176: *for the datagrid

app/Models/MapMarker.php
```diff
@@ -74,6 +75,11 @@ class MapMarker extends Model
         'polygon_style',
     ];
 
+    protected $sortable = [
+        'name',
+        'entity_id',
```
ilestis, line 80: Also add type, icon, group.name and visibility as sortable options

app/Renderers/Layouts/Map/Marker.php
```diff
@@ -0,0 +1,85 @@
...
+     * @return array[]
+     */
+    public function columns(): array
+    {
+        $columns = [
+            'name' => [
+                'key' => 'name',
+                'label' => 'crud.fields.name',
+                'render' => function ($model) {
+                    return $model->tooltipedLink($model->markerTitle());
```
ilestis, line 21: tooltipedLink is used for links with tooltips and an ajax request on hover, so this name isn't consistent with the rest of the app. Also, using markerTitle() makes it that the order by name makes no sense 
image

app/Renderers/Layouts/Map/Marker.php
```diff
@@ -0,0 +1,85 @@
...
+                },
+            ],
+            'entity_id' => [
+                'label' => 'crud.fields.entity',
+                'render' => function ($model) {
+                    if ($model->entity_id) {
+                        return $model->tooltipedLink($model->entity->name);
```
ilestis, line 28: Missing a check that entity isn't empty, otherwise for a non-admin this page will crash if entity_id is set, but the user doesn't have permission to see the marker.

Can be simplified by
`return $model->entity?->tooltipedLink()`, that way the tooltips also work

app/Renderers/Layouts/Map/Marker.php
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                    return '';
+                }
+            ],
+            'groups' => [
+                'label' => 'maps/markers.fields.group',
+                'render' => function ($model) {
+                    if ($model->group) {
```
ilestis, line 36: Can be simplified to `return $model->group?->name`

resources/views/maps/markers/index.blade.php
```diff
@@ -0,0 +1,44 @@
...
+    @endcan
+@endsection
+
+@section('content')
+    <div class="entity-grid">
+        @include('entities.components.header', [
+            'model' => $model,
+            'breadcrumb' => [
+                ['url' => Breadcrumb::index('maps'), 'label' => __('maps.index.title')],
+                null
```
ilestis, line 33: Missing the "Markers" breadcrumb. Same for Layers and Groups. See the Abilities or INventory subpage for what I mean
image

ilestis (changes requested):
A few tweaks, good job!

resources/views/maps/form/_markers.blade.php
```diff
@@ -150,37 +150,33 @@ class="fa-solid fa-map"></i> {{ __('maps.actions.explore') }}</a>
...
-                        @include('maps.markers._form', ['model' => null, 'map' => $model, 'activeTab' => 1, 'dropdownParent' => '#marker-modal'])
-
-                        <div class="form-group">
-                            <div class="pull-left">
-                                @include('partials.footer_cancel', ['ajax' => 1])
-                            </div>
-                            <div class="pull-right">
-                                @include('maps.markers._actions')
-                            </div>
+                <div class="panel-heading">
```
ilestis, line 153: image
`panel-*` classes are still wrong when creating a marker in an modal, it should be "modal-body" and "modal-footer".

ilestis (changes requested):
Small fixes

PR #520, opened 2022-10-24
The new subpages are now available on the dropdowns in the list of th…

spitfire305:
…e campaign's maps

### DIFF
--- a/app/Models/MiscModel.php
+++ b/app/Models/MiscModel.php
@@ -585,6 +585,23 @@ abstract class MiscModel extends Model
                 </a>
             </li>';
             }
+            if ($campaign->enabled('maps') && $this->entityTypeId() == config('entities.ids.map')) {
+                $actions[] = '<li>
+                <a href="' . route('maps.map_layers.index', $this->id) . '" class="dropdown-item datagrid-dropdown-item" data-name="layers">
+                    <i class="fa-solid fa-layer-group" aria-hidden="true"></i> ' . __('maps.panels.layers') . '
+                </a>
+            </li>';
+                $actions[] = '<li>
+                <a href="' . route('maps.map_groups.index', $this->id) . '" class="dropdown-item datagrid-dropdown-item" data-name="groups">
+                    <i class="fa-solid fa-map-signs" aria-hidden="true"></i> ' . __('maps.panels.groups') . '
+                </a>
+            </li>';
+                $actions[] = '<li>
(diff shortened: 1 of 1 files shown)


app/Models/MiscModel.php
```diff
@@ -585,6 +585,23 @@ public function datagridActions(Campaign $campaign): array
                 </a>
             </li>';
             }
+            if ($campaign->enabled('maps') && $this->entityTypeId() == config('entities.ids.map')) {
```
ilestis, line 588: This code needs to be on the Map model, since MiscModel doesn't need to know elements specific to Maps

ilestis (changes requested):
Move the code to the Map model to keep MiscModel lean and clean

app/Models/Map.php
```diff
@@ -720,4 +720,69 @@ public function hasDistanceUnit(): bool
...
+     * @param Campaign $campaign
+     * @return string[]
+     * @throws Exception
+     */
+    public function datagridActions(Campaign $campaign): array
+    {
+        $actions = [];
+
+        // Relations & Inventory
+        if (!isset($this->hasRelations)) {
```
ilestis, line 734: This looks like copy-pasted code. You can simply call `parent::datagridActions()` to call the method from MiscModel :)

ilestis (changes requested):
Remove copy-pasted code :)

app/Models/Map.php
```diff
@@ -728,60 +728,28 @@ public function hasDistanceUnit(): bool
...
-            if ($campaign->enabled('abilities') && $this->entityTypeId() != config('entities.ids.ability')) {
-                $actions[] = '<li>
-                <a href="' . route('entities.entity_abilities.index', $this->entity) . '" class="dropdown-item datagrid-dropdown-item" data-name="abilities">
-                    <i class="ra ra-fire-symbol" aria-hidden="true"></i> ' . __('crud.tabs.abilities') . '
-                </a>
-            </li>';
-            }
-            if ($campaign->enabled('maps') && $this->entityTypeId() == config('entities.ids.map')) {
-                $actions[] = '<li>
+        if ($campaign->enabled('maps') && $this->entityTypeId() == config('entities.ids.map') && auth()->check() && auth()->user()->can('update', $this)) {
```
ilestis, line 734: We know we're on a map, so the map module is always enabled, and the entity type is already a map, so we only need to check for the permission to update the entity ;-)

ilestis (changes requested):
Tiny last change to keep the code clean

ilestis (approved):

PR #525, opened 2022-10-26
Added support for bulk editing map markers

spitfire305:


resources/views/layouts/datagrid/bulks/_map-marker.blade.php
```diff
@@ -1,3 +1,83 @@
...
+    0 => __('general.no'),
+    1 => __('general.yes'),
+];
+
+$groups = $model->groupOptions();
+$groups[-1] = __('crud.filters.options.none');
+?>
+<div class="form-group">
+    <label for="icon">{{ __('locations.map.points.fields.icon') }}</label>
+    {!! Form::select('icon', $iconOptions, \App\Facades\FormCopy::field('icon')->string(), ['class' => 'form-control', 'id' => 'icon']) !!}
```
ilestis, line 34: We don't need to call FormCopy in this place, since we aren't copying a marker :)

ilestis (changes requested):
Tiny tweaks, copy-pasting cleanup

ilestis (approved):

PR #526, opened 2022-10-27
Added the ability for users to filter templates.

spitfire305:


app/Models/Concerns/HasFilters.php
```diff
@@ -375,6 +378,23 @@ protected function filterHasImage(Builder $query, string $value = null): void
...
+     * @param Builder $query
+     * @param string|null $value
+     * @return void
+     */
+    protected function filterTemplate(Builder $query, string $value = null): void
+    {
+        $query = $this->joinEntity($query);
+
+        if ($value) {
+            $query->whereNotNull('e.is_template');
```
ilestis, line 392: An entity that was a template and no longer is is appearing with this query.

app/Models/Concerns/HasFilters.php
```diff
@@ -375,6 +378,23 @@ protected function filterHasImage(Builder $query, string $value = null): void
...
+        } else {
+            $query->whereNull('e.is_template');
```
ilestis, line 394: An entity that was a template and no longer is does''t appear with this query.

ilestis (changes requested):
Wrong SQL for no longer templates

app/Models/Concerns/HasFilters.php
```diff
@@ -389,9 +389,9 @@ protected function filterTemplate(Builder $query, string $value = null): void
         $query = $this->joinEntity($query);
 
         if ($value) {
-            $query->whereNotNull('e.is_template');
+            $query->where('e.is_template', 1);
         } else {
-            $query->whereNull('e.is_template');
+            $query->whereNull('e.is_template')->orWhere('e.is_template', '<>', 1);
```
ilestis, line 394: When I do a filter for name `Adam` + is not template, it's not showing me the results as expected.
image

```sql
select distinct ....
 from `characters`
 left join `entities` as `e` on `e`.`entity_id` = `characters`.`id` and `e`.`type_id` = 1 and e.campaign_id = characters.campaign_id 
where (`characters`.`name` like '%Adam%' and `e`.`is_template` is null or `e`.`is_template` <> 1) 
...
```
It's not separating the two is_template queries into a single () statement

ilestis (changes requested):
Doesn't work when combined with other filters

ilestis (approved):

PR #527, opened 2022-10-28
Disabled tooltip popup on mouseover for dashboiard map markers

spitfire305:


app/Models/MapMarker.php
```diff
@@ -318,7 +321,7 @@ protected function popup(): string|null
                 ' . $body . '
             </div>`)
             .on(`mouseover`, function (ev) {
```
ilestis, line 323: Feel like having an if around the whole `on()` would be cleaner, rather than having an empty `mouseover` :)

app/Models/MapMarker.php
```diff
@@ -499,9 +502,12 @@ public function editing(): self
      * Set the current mode to exploring the map
      * @return $this
      */
-    public function exploring(): self
+    public function exploring($popup = true): self
```
ilestis, line 505: Missing type casting + param documentation

ilestis (changes requested):
Tweak to make code a bit cleaner

ilestis (approved):

PR #528, opened 2022-10-31
Save button for marker editing now redirects the user to the correct …

spitfire305:
…page

### DIFF
--- a/app/Http/Controllers/Maps/MapMarkerController.php
+++ b/app/Http/Controllers/Maps/MapMarkerController.php
@@ -135,7 +135,7 @@ class MapMarkerController extends Controller
      * @return \Illuminate\Contracts\Foundation\Application|\Illuminate\Contracts\View\Factory|\Illuminate\View\View
      * @throws \Illuminate\Auth\Access\AuthorizationException
      */
-    public function edit(Map $map, MapMarker $mapMarker)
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Maps/MapMarkerController.php
```diff
@@ -135,7 +135,7 @@ public function store(Map $map, StoreMapMarker $request)
      * @return \Illuminate\Contracts\Foundation\Application|\Illuminate\Contracts\View\Factory|\Illuminate\View\View
      * @throws \Illuminate\Auth\Access\AuthorizationException
      */
-    public function edit(Map $map, MapMarker $mapMarker)
+    public function edit(Map $map, MapMarker $mapMarker, string $from = '' )
```
ilestis, line 138: It's cleaner to just use request() than passing params in the function call

app/Http/Controllers/Maps/MapMarkerController.php
```diff
@@ -147,6 +147,13 @@ public function edit(Map $map, MapMarker $mapMarker)
         $includeMap = true;
         $activeTab = $mapMarker->shape_id;
 
+        if (request()->from == 'explore') {
```
ilestis, line 150: Especially since you ignore the from here :P

ilestis (changes requested):

PR #529, opened 2022-10-31
Lookup bar now shows an extra result redirecting to the explore page …

spitfire305:
…for maps that can be explored

### DIFF
--- a/app/Services/SearchService.php
+++ b/app/Services/SearchService.php
@@ -259,6 +259,21 @@ class SearchService
                 'alias_id' => $model->alias_id, // @phpstan-ignore-line
                 'advanced_mention' => Mentions::advancedMentionHelper($model->name),
             ];
+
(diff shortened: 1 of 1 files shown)


app/Services/SearchService.php
```diff
@@ -259,6 +259,21 @@ public function find()
                 'alias_id' => $model->alias_id, // @phpstan-ignore-line
                 'advanced_mention' => Mentions::advancedMentionHelper($model->name),
             ];
+
+            //If the result is a map, also add its explore page as a result.
+            if ($model->type() == 'map' && $model->child->explorable()) {
```
ilestis, line 264: It's now showing up in the @mention system, which isn't needed. It's the same endpoint, but we have a param that separates both.
image

ilestis (changes requested):

PR #532, opened 2022-11-03
Edit warning when editing entities

spitfire305:


app/Http/Controllers/Entity/EditingController.php
```diff
@@ -6,7 +6,12 @@
 
 use App\Http\Controllers\Controller;
 use App\Models\Entity;
+use App\Models\EntityNote;
+use App\Models\Campaign;
+use App\Models\QuestElement;
+use App\Models\TimelineElement;
 use App\Services\Entity\MultiEditingService;
+use Ramsey\Uuid\Type\Time;
```
ilestis, line 14: Didn't see where this is being used

app/Http/Controllers/Entity/EditingController.php
```diff
@@ -6,7 +6,12 @@
...
 
 class EditingController extends Controller
```
ilestis, line 16: Since this isn't anymore limited to just Entities, we can move the namespace and route names

app/Http/Controllers/Entity/EditingController.php
```diff
@@ -35,18 +40,121 @@ public function confirm(Entity $entity)
...
+            ->confirm();
+
+        return response()->json(['success' => true]);
+    }
+
+    public function confirmPost(Entity $entity, EntityNote $entityNote)
+    {
+        $this->authorize('entity-note', [$entity->child, 'edit', $entityNote]);
+
+        $this->service
```
ilestis, line 59: Lots of copy-pasted code here that we can remove with some refactoring :)

app/Http/Controllers/Entity/EditingController.php
```diff
@@ -35,18 +40,121 @@ public function confirm(Entity $entity)
...
+            ->json([
+                'success' => true
+            ]);
+    }
+
+    public function keepAlivePost(Entity $entity, EntityNote $entityNote)
+    {
+        $this->authorize('entity-note', [$entity->child, 'edit', $entityNote]);
+
+        $this->service->model($entityNote)
```
ilestis, line 123: Same for the keepalive codes, lots of copy-pasting

app/Models/Campaign.php
```diff
@@ -580,4 +581,14 @@ public function follower(): int
     {
         return (int) $this->follower;
     }
+
+    /**
+     * @return mixed
+     */
+    public function users()
```
ilestis, line 588: This method name is very confusing to me, since you'd expect the campaign users to be its members. Also the other relations are all placed in the CampaignRelations model, to keep the campaign model clean.

app/Models/EntityNote.php
```diff
@@ -167,4 +169,14 @@ public function collapsed(): bool
     {
         return Arr::get($this->settings, 'collapsed', false);
     }
+
+    /**
+     * @return mixed
+     */
+    public function users()
```
ilestis, line 176: Same here, the name users is confusing to someone reading the code for the first time, especially since there are no comments in the code. editingUsers() would make more sense.

app/Models/QuestElement.php
```diff
@@ -104,4 +106,14 @@ public function hasEntity(): bool
     {
         return false;
     }
+
+    /**
+     * @return mixed
+     */
+    public function users()
```
ilestis, line 113: Same here, editingUsers would make more sense

app/Models/TimelineElement.php
```diff
@@ -167,4 +169,14 @@ public function collapsed(): bool
     {
         return $this->is_collapsed;
     }
+
+    /**
+     * @return mixed
+     */
+    public function users()
```
ilestis, line 176: Same

app/Services/Entity/MultiEditingService.php
```diff
@@ -73,7 +78,17 @@ public function isEditing(): bool
     public function edit(): self
     {
         $model = new EntityUser();
-        $model->entity_id = $this->entity->id;
+        if ($this->model->getTable() == 'entity_notes') {
```
ilestis, line 81: If we change the table name, all this code will need to be manually updated. Instead you can use `$this->model instanceof EntityNote` to check the object's class.

app/Services/Entity/MultiEditingService.php
```diff
@@ -86,9 +101,21 @@ public function edit(): self
      */
     public function finish(): self
     {
+        if ($this->model->getTable() == 'entity_notes') {
```
ilestis, line 104: Same here, use `instanceof` instead of testing the table name which can be changed

resources/views/entities/pages/entity-notes/edit.blade.php
```diff
@@ -40,4 +44,39 @@
     @parent
     {!! Form::open(['method' => 'DELETE', 'route' => ['entities.entity_notes.destroy', 'entity' => $entity, 'entity_note' => $model], 'style' => 'display:inline', 'id' => 'delete-form-note-' . $model->id]) !!}
     {!! Form::close() !!}
+
+    @if(!empty($editingUsers) && !empty($model))
```
ilestis, line 48: We can just define all this HTML only once and use an include to avoid copy-pasting :)

ilestis (changes requested):
Good job! Next up: removing all the copy-pasting so that adding this feature in the future to more elements is easier 💪🏻

ilestis:
The warning popup always talks about `posts` instead of using the element's type
image

resources/views/cruds/forms/edit_warning.blade.php
```diff
@@ -0,0 +1,49 @@
+<?php
+    if ($model instanceof \App\Models\EntityNote) {
```
ilestis, line 2: Something here isn't working

ilestis (changes requested):

ilestis (approved):

PR #533, opened 2022-11-04
Added new user logs for when emails are sent.

spitfire305:


app/Jobs/Emails/SubscriptionDeletedEmailJob.php
```diff
@@ -56,6 +57,6 @@ public function handle()
             ->send(
                 new DeletedSubscriptionMail($user)
             );
-
+            $user->log(UserLog::TYPE_SUB_CANCEL_AUTO);
```
ilestis, line 60: When a user manually cancels, it now logs it twice

ilestis (changes requested):

PR #537, opened 2022-11-08
Users can now unmirror relations from the connections page of an entity.

spitfire305:


app/Http/Controllers/Entity/RelationController.php
```diff
@@ -204,7 +204,13 @@ public function edit(Entity $entity, Relation $relation)
     public function update(StoreRelation $request, Entity $entity, Relation $relation)
     {
         $this->authorize('update', $entity->child);
-        $data = $request->only(['target_id', 'attitude', 'relation', 'colour', 'is_star', 'two_way', 'visibility_id']);
+        $data = $request->only(['target_id', 'attitude', 'relation', 'colour', 'is_star', 'two_way', 'visibility_id', 'unmirror']);
```
ilestis, line 207: Seems cleaner to skip adding `unmirror`, and simply test `if ($request->has('unmirror')`, and skip the `unset()`

ilestis (changes requested):
Small tweak to make the code cleaner

PR #543, opened 2022-11-24
Imported character sheets now render on the proper language for the u…

spitfire305:
…ser.

### DIFF
--- a/app/Models/PluginVersion.php
+++ b/app/Models/PluginVersion.php
@@ -6,6 +6,8 @@ use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Translation\Translator;
+use Illuminate\Contracts\Translation\Loader;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


app/Models/PluginVersion.php
```diff
@@ -143,13 +154,31 @@ protected function renderBlade(Entity $entity)
 
         $html = str_replace([
             '@php', '@dd', '@inject', '@yield', '@section', '@auth', '@guest', '@env', '@once', '@push', '@csrf',
-            '@include', '\Illuminate\\',
+            '@include', '\Illuminate\\', "@i18n('"
```
ilestis, line 157: This block is for removing blacklisted functions, so wrong place to add this :)

app/Models/PluginVersion.php
```diff
@@ -143,13 +154,31 @@ protected function renderBlade(Entity $entity)
...
-            '', '', '', '', '', '', '', '', '', '', ''
+            '', '', '', '', '', '', '', '', '', '', '', '', '', "trans('*."
         ], $html);
         $html = preg_replace('`dd\((.*?)\)`i', '', $html);
         $html = preg_replace('`config\((.*?)\)`i', '', $html);
+        $locale = auth()->user()->locale;
+        $lines = [];
+
+        //Replace translation keys of existing translations
+        foreach ($this->getTranslationsAttribute() as $translation) {
```
ilestis, line 167: You can replace these whole two blocks with a single `preg_replace_content` like it's done for `@liveAttribute` and 3 lines :-)

ilestis (changes requested):
Can be refactored into 3 lines of code

app/Models/PluginVersion.php
spitfire305: `preg_replace_callback` doesnt seem to be working, do I leave the current implementation while its fixed?

PR #546, opened 2022-11-28
Marketplace: allow hidden attributes

spitfire305:


resources/views/cruds/forms/attributes/_buttons.blade.php
```diff
@@ -37,7 +37,12 @@
         @endif
     </ul>
 </div>
-
+@if ($entity->attributes()->where('is_hidden', '0')->get()->has('0'))
```
ilestis, line 40: When creating a new entity, I have a crash because of this line. Same when editing an entity.

ilestis (changes requested):

ilestis (approved):
Nice!

PR #547, opened 2022-11-29
Added the data-tags attribute to entity abilities.

spitfire305:


app/Services/Entity/AbilityService.php
```diff
@@ -51,17 +51,27 @@ public function abilities(): array
...
             ->defaultOrder()
             ->get();
         /** @var EntityAbility $ability */
-        foreach ($abilities as $ability) {
+        foreach ($abilities as $key => $ability) {
             // Can't read the ability? skip
             if (empty($ability->ability) || empty($ability->ability->entity)) {
                 continue;
             }
+            $class = '';
```
ilestis, line 65: Why not have this whole block in the `format()` method, since that's also where the ability is transformed into json?

ilestis (changes requested):

ilestis (approved):
Perfect :)

PR #550, opened 2022-12-02
Marketplace allow hidden attribute

spitfire305:


app/Http/Controllers/Abilities/AbilityReorderController.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace App\Http\Controllers\Abilities;
```
ilestis, line 3: I think it would make more sense to put this in the Entity namespace, since it's related to an entity's abilities :)

routes/campaign.php
```diff
@@ -23,6 +23,11 @@
 Route::get('/abilities/{ability}/entity-add', 'AbilityController@entityAdd')->name('abilities.entity-add');
 Route::post('/abilities/{ability}/entity-add', 'AbilityController@entityStore')->name('abilities.entity-add.save');
 
+//Ability reorder
+Route::get('/entity/{entity}/abilities/reorder', [\App\Http\Controllers\Abilities\AbilityReorderController::class, 'index'])
```
ilestis, line 27: A `entities.entity_abilities.reorder` route name would also make more sense in my opinion, keep it similar to the other entity ability route names. (and place them in the same area in the file)

app/Http/Controllers/Entity/AbilityController.php
```diff
@@ -166,7 +166,7 @@ public function update(StoreEntityAbility $request, Entity $entity, EntityAbilit
     {
         $this->authorize('update', $entity->child);
 
-        $data = $request->only(['ability_id', 'visibility_id', 'note', 'position']);
+        $data = $request->only(['ability_id', 'visibility_id', 'note']);
```
ilestis, line 169: Unrelated, but the api endpoint no longer works on my local version
image

app/Models/MiscModel.php
```diff
@@ -333,6 +333,13 @@ public function menuItems(array $items = []): array
                 'entity' => true,
                 'icon' => 'ra ra-fire-symbol',
             ];
+            if (auth()->check() && auth()->user()->can('update', $this)) {
```
ilestis, line 336: You can remove the link from the sidebar, there are already enough things there.

resources/views/entities/pages/abilities/index.blade.php
```diff
@@ -17,6 +17,12 @@
                 <i class="fa-solid fa-redo"></i> <span class="hidden-xs hidden-sm">{{ __('entities/abilities.actions.reset') }}</span>
                 <span class="visible-xs visible-sm">{{ __('crud.actions.reset') }}</span>
             </a>
+            <a href="{{ route('abilities.reorder', $entity) }}" class="btn btn-sm btn-default">
+                <i class="fa-solid fa-sort"></i> 
+                <span class="visible-xs visible-sm">
```
ilestis, line 22: Missing the text on desktop, it's just an icon?

lang/en/abilities.php
```diff
@@ -39,6 +39,12 @@
         'tabs'  => [
             'abilities' => 'Sub abilities',
             'entities'  => 'Entities',
+            'reorder'   => 'Reorder Abilities',
```
ilestis, line 42: You can move all of these translations to `entities/abilities.php` instead, since it's related to entity abilities, and note the ability objects directly.

resources/views/entities/pages/abilities/reorder/_reorder.blade.php
```diff
@@ -0,0 +1,50 @@
...
+    <div class="box-header">
+        <h3 class="box-title">
+            {{ __('abilities.reorder.title') }}
+        </h3>
+    </div>
+    <div class="box-body">
+
+    @foreach($parents as $key => $parent)
+
+        <h3 class="box-title">
```
ilestis, line 16: `box-title` only works in a `box-header` parent div. Here the class does nothing, so it can be removed.

ilestis (changes requested):

resources/views/entities/pages/abilities/index.blade.php
spitfire305: Just an icon

ilestis (approved):

PR #554, opened 2022-12-05
Changed field type of the quantity field for the add new item to inve…

spitfire305:
…ntory form.

### DIFF
--- a/resources/views/entities/pages/inventory/_form.blade.php
+++ b/resources/views/entities/pages/inventory/_form.blade.php
@@ -34,7 +34,7 @@
     <div class="col-md-6">
         <div class="form-group required">
             <label>{{ __('entities/inventories.fields.amount') }}</label>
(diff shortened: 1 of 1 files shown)


resources/views/entities/pages/inventory/_form.blade.php
```diff
@@ -34,7 +34,7 @@
     <div class="col-md-6">
         <div class="form-group required">
             <label>{{ __('entities/inventories.fields.amount') }}</label>
-            {!! Form::text('amount', (empty($inventory) ? 1 : null), ['placeholder' => __('entities/inventories.placeholders.amount'), 'class' => 'form-control', 'maxlength' => 191]) !!}
+            {!! Form::number('amount', (empty($inventory) ? 1 : null), ['class' => 'form-control', 'maxlength' => 191]) !!}
```
ilestis, line 37: maxlength doesn't make sense on a numeric, and it's probably worth having a "min" and "max" value based on the database field (hardcoded in the view like for other number fields) :)

ilestis (changes requested):

ilestis (approved):

PR #555, opened 2022-12-05
Now two arrows show for mirrored relations with different names.

spitfire305:


app/Services/Entity/EntityRelationService.php
```diff
@@ -243,7 +243,7 @@ protected function addRelations(Entity $entity): self
                 ])
             ];
 
-            if ($relation->isMirrored()) {
+            if ($relation->isMirrored() && $relation->relation == $relation->mirror->relation) {
```
ilestis, line 246: This has added 30% more db calls on a test entity on my end, missing a with() somewhere :)

ilestis: Also, I think the "unmirror" check is broken. Whenever I edit a relation that is mirrored, even without changing anything, the relation is no longer mirrored.

ilestis (changes requested):

PR #556, opened 2022-12-07
Users can now assing filters to quicklinks for entities

spitfire305:


resources/assets/js/quick-links.js
```diff
@@ -20,3 +21,18 @@ function initQuickLinksForm() {
         });
     });
 }
+
+function showFilterField() {
+    selector = $('#entity-selector');
```
ilestis, line 26: Where is this variable defined? I'm getting js errors in my editor. Variables need to be declared as either var (general) or let (local) in javascript :)
image

resources/assets/js/quick-links.js
```diff
@@ -20,3 +21,18 @@ function initQuickLinksForm() {
...
+    if (selector.val() != '') {
```
ilestis, line 27: Missing a test to make sure the field exists before testing it's value, to avoid having scripts running when not needed, slowing down the browser.

app/Models/MenuLink.php
```diff
@@ -219,7 +219,12 @@ protected function getEntityRoute(): string
                 $route = $menuRoute;
             }
         }
-        return route($route, $this->getRouteParams());
+        $filter = '';
```
ilestis, line 222: Why not add this check in getRouteParams()? That's literally what the function does, add params to the route, and is a lot cleaner than adding the params as a string, and doesn't break when switching between `?` and `&` for the params to be added.

public/js/app.js
```diff
@@ -4771,9 +4771,11 @@ function initQuickLinksForm() {
 }
 
 function showFilterField() {
-  selector = $('#entity-selector');
+  var selector = $('#entity-selector');
```
ilestis, line 4774: Since selector is only used in this function, `let` is the correct declarator, since we have a `let selector` in other functions in the same script.

ilestis (changes requested):

app/Models/MenuLink.php
```diff
@@ -169,6 +169,11 @@ public function getRouteParams()
                 $parameters['all_members'] = 1;
             }
         }
+
+        if ($this->menu && isset($this->options['subview_filter'])) {
```
ilestis, line 173: The previous if already checks if it's going to a submenu, so you can add the second section of the if in the previous if block to make the code easier to read ;-)
```php
if (!empty($this->menu)) {
   if ($this->menu == 'all-members') {
   }
   if (isset($this->options['subview_filter'])) {
   }
}
```

ilestis (changes requested):
Small code readability improvement

ilestis (approved):

PR #558, opened 2022-12-08
Apply attribute template workflow

spitfire305:


resources/views/entities/pages/attribute-templates/_form.blade.php
```diff
@@ -22,15 +21,11 @@
...
-
-        <div class="my-5 text-center">
-            <button type="button" class="btn btn-default mr-5 rounded-full px-8" data-dismiss="modal">
-                {{ __('crud.cancel') }}
-            </button>
-
-            <button class="btn btn-success ml-5 rounded-full px-8">
-                {{ __('crud.actions.apply') }}
-            </button>
+        <div class="pull-right">
```
ilestis, line 24: Please keep the old "centered buttons" look to keep it consistent with the other round-button popups 
image
Hint: you'll want to add `inline-block` as a class to the `submit-group` div.

ilestis (changes requested):
Keep the UI consistent with other interfaces with rounded buttons

ilestis (approved):

PR #561, opened 2022-12-12
1 month long calendars can be created now.

spitfire305:


resources/views/calendars/_calendar.blade.php
```diff
@@ -2,12 +2,6 @@
 /** @var \App\Renderers\CalendarRenderer $renderer
  * @var \App\Models\Calendar $model
  */
-if ($model->missingDetails()): ?>
```
ilestis, line 5: If I create a calendar in the quick creator with 0 months, showing the calendar now crashes without this check

ilestis (changes requested):

ilestis (approved):

PR #562, opened 2022-12-12
Calendars now mark the start and end of an event when rendered.

spitfire305:


ilestis (changes requested):
If an event starts on the 30th of december, but ends in the 3rd on january, it shows (start) but not `(end)`
image

app/Renderers/CalendarRenderer.php
```diff
@@ -856,6 +868,24 @@ protected function parseReminders(Collection $reminders): void
                 $this->addMultidayEvent($event, $date);
             }
         }
+
+        $start = [];
+        $end = [];
+        foreach ($this->events as $date => $events) {
```
ilestis, line 874: It doesn't work for recurring events :) 
image

I think this whole code should be in the `addMultliEvent()` function to properly loop on each occurance of an event, not just the first one (an event can appear multiple times in the same month if it's set to happen on a new moon)

ilestis (changes requested):
Don't work on recurring events :D

app/Renderers/CalendarRenderer.php
```diff
@@ -906,6 +897,21 @@ protected function addMultidayEvent(EntityEvent $event, string $date, bool $recu
             }
             $this->events[$extraDate][] = $event;
         }
+        $start = [];
+        $end = [];
+        foreach ($this->events as $date => $events) {
```
ilestis, line 902: I'm confused why you are looping on $this->events when this function is already called on each event that is added. This creates a n+1 loop for each event displayed, resetting $this->eventEnd on each event drawn on the calendar.

ilestis (changes requested):
The code can still be improved to avoid an exponential loop :)

PR #563, opened 2022-12-14
Arrows now show properly on mirrorer relations

spitfire305:


app/Services/Entity/EntityRelationService.php
```diff
@@ -233,7 +233,7 @@ protected function addRelations(Entity $entity): self
                 'attitude' => $relation->attitude,
                 'type' => 'entity-relation',
                 'is_mirrored' => $relation->isMirrored(),
-                'shape' => $relation->isMirrored() ? 'none' : 'triangle',
+                'shape' => $relation->isMirrored() && $relation->relation == $relation->mirror->relation ? 'none' : 'triangle',
```
ilestis, line 236: Missing a permission check if the mirrored relation is private from the user to view, causing a page crash `$relation->isMirrored() && $relation->mirror && $relation->relation == $relation->mirror->relation` :)

ilestis (changes requested):
Permission check to avoid the page crashing

ilestis (approved):

PR #565, opened 2022-12-16
Added the option for users to skip year 0 on calendars

spitfire305:


ilestis (changes requested):
Fun edge case, on the calendar dashboard widget, the "days ago" is wrong if the calendar skips year zero, is on year one, and there is an event on year -1 :) 
image

ilestis (approved):

ilestis (approved):

PR #566, opened 2022-12-19
Added the option for users to hide some tags from entity headers and …

spitfire305:
…tooltips.

### DIFF
--- a/app/Models/Relations/EntityRelations.php
+++ b/app/Models/Relations/EntityRelations.php
@@ -238,7 +238,7 @@ trait EntityRelations
     public function tagsWithEntity()
     {
         if ($this->tagsWithEntity === false) {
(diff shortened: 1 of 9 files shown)


app/Models/Relations/EntityRelations.php
```diff
@@ -238,7 +238,7 @@ public function tags()
     public function tagsWithEntity()
     {
         if ($this->tagsWithEntity === false) {
-            $this->tagsWithEntity = $this->tags()->with('entity')->get();
+            $this->tagsWithEntity = $this->tags()->with('entity')->where('is_hidden', '=', '0')->get();
```
ilestis, line 241: Since we're doing a `where()` on the db, we need to index the field to avoid bad performances.

ilestis (changes requested):
Missing an index

ilestis (approved):

PR #570, opened 2022-12-30
Ids are automatically added to headings.

spitfire305:


app/Models/MiscModel.php
```diff
@@ -429,6 +429,7 @@ public function entityTypeSuggestion(int $take = 20): array
      */
     public function entry()
     {
+        //changes here?
```
ilestis, line 432: Are you sure? :P

ilestis (changes requested):

app/Models/MiscModel.php
spitfire305: Turns out changes not there

ilestis (approved):

PR #571, opened 2022-12-31
Printing now works properly

spitfire305:


resources/views/layouts/print.blade.php
```diff
@@ -4,7 +4,7 @@
  * @var \App\Models\MiscModel $miscModel
  */
 $campaign = CampaignLocalization::getCampaign();
-$themeOverride = request()->get('_theme');
+$themeOverride = request()->get('_theme') ? : "base";
```
ilestis, line 7: The second parameter of `->get()` is the default, so you can simplify the code even more :)

ilestis (changes requested):

ilestis (approved):

PR #572, opened 2023-01-03
Discord auth tokens are now automatically renewed each day.

spitfire305:


app/Services/DiscordService.php
```diff
@@ -158,6 +159,13 @@ public function refresh()
         ];
         $content = $this->post('oauth2/token', $body);
         $this->saveUserApp($content);
+        $log = 'Renewed user #' . $this->user->id . ' Discord auth token.';
+        $this->logs[] = $log;
+
+        JobLog::create([
```
ilestis, line 165: This is going to create one JobLog when the user renews themselves (this not on a job), and also create a log for every user (potentially 100s a day), please refactor to only do a single log and just for the command :) Also only needs to create the log if the local install has jog logs (not available for self-hosted kanka)

ilestis (changes requested):

app/Services/DiscordService.php
```diff
@@ -310,4 +314,20 @@ public function logs(): array
...
+     * @return void
+     */
+    public function log()
+    {
+        if (!config('app.log_jobs')) {
+            return;
+        }
+
+        JobLog::create([
+            'name' => 'subscriptions:end',
```
ilestis, line 329: The `name` of the job seems wrong ;-)

app/Console/Commands/RegenerateDiscordToken.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function handle()
+    {
+        //$userID = $this->argument('user');
+
+        /** @var User $user */
+        $tokens = UserApp::select(['id', 'user_id', 'access_token', 'refresh_token', 'expires_at', 'updated_at'])
+            ->with('user')
+            ->where('app', '=', 'discord')
+            //->where('expires_at', '<=', Carbon::now()->toDateString())
```
ilestis, line 54: I don't think this should be commented out 🤔

ilestis (changes requested):
Small fixes

ilestis (approved):

PR #574, opened 2023-01-03
Users can now unlink discord accounts even when the tokens are invalid

spitfire305:


app/Services/DiscordService.php
```diff
@@ -246,7 +246,11 @@ public function remove(): self
...
         // Remove any roles the user might have had
-        $this->removeRoles();
+        try {
+            $this->removeRoles();
+        } catch (Exception $e) {
+            $this->app->delete();
+        }
 
         // Delete the discord app
         $this->app->delete();
```
ilestis, line 256: In case of an error, you're deleting the discord app twice, which doesn't sound bug-free. The `catch` should be silent and not do anything, so it continues even on an error.

ilestis (changes requested):
Small refactoring

ilestis (approved):

PR #575, opened 2023-01-04
Users can now create tags while bulk editing or quick creating entities

spitfire305:


ilestis (approved):
It's not clean code (does 3 db queries per tag) but that's due to the EntityObserver. Will need to be refactored one day 🥲

PR #578, opened 2023-01-06
Third section of the sidebar automatically sorts alphabetically accor…

spitfire305:
…ding to the users language

### DIFF
--- a/resources/views/entities/components/menu.blade.php
+++ b/resources/views/entities/components/menu.blade.php
@@ -2,6 +2,15 @@
 @php $modelMenuItems = $model->menuItems(); @endphp
 <div class="hidden-xs">
 @foreach ($modelMenuItems as $section => $menuItems)
+    @if($section == 2)
+        @php
+            $items = array_column($menuItems, 'name');
+            foreach ($items as $key => $item) {
+                $items[$key] = __($item);
+            }
+            array_multisort($items, SORT_ASC, $menuItems);
+        @endphp
+    @endif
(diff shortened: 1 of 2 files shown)


resources/views/entities/components/menu.blade.php
```diff
@@ -2,6 +2,15 @@
 @php $modelMenuItems = $model->menuItems(); @endphp
 <div class="hidden-xs">
 @foreach ($modelMenuItems as $section => $menuItems)
+    @if($section == 2)
```
ilestis, line 5: Please mode the logic to the Model and not in the view, so that we can test the code in the future :)

resources/views/entities/components/menu.blade.php
```diff
@@ -2,6 +2,15 @@
...
+        @php
+            $items = array_column($menuItems, 'name');
+            foreach ($items as $key => $item) {
+                $items[$key] = __($item);
+            }
+            array_multisort($items, SORT_ASC, $menuItems);
```
ilestis, line 11: SORT_ASC doesn't work with accented characters (Russian, French). 

1. Switch to French
2. open `lang/fr/crud.php` and find `tabs.inventory` and replace it with `Înventaire` (accented uppercase letter)
3. Refresh the browser, see `Înventaire` all at the end, when it should be in second position.

See how it's done in the quick creator's entity type list to sort alphabetically for international languages :)

ilestis (changes requested):
Doesn't work with accented first letters

ilestis (approved):
👌🏻

PR #580, opened 2023-01-09
Maintenance mode warning part 2

spitfire305:


resources/assets/js/crud.js
```diff
@@ -347,7 +347,10 @@ function registerUnsavedChanges() {
  * When the entity form is submitted, we want to ajax validate the request first
  */
 function registerEntityFormSubmit() {
-    $('#entity-form').submit(function (e) {
+    entityFormActions =$('form[data-unload="1"]');
```
ilestis, line 350: This variable is declared but I don't see it ever users?

ilestis (changes requested):
I think there is some leftover testing variable in the js file

ilestis (approved):

PR #582, opened 2023-01-10
New Sub: Renewed after cancelling

spitfire305:


ilestis (changes requested):
Duplicate code

app/Models/Relations/UserRelations.php
```diff
@@ -166,4 +166,13 @@ public function bragiLogs()
...
         return $this->hasMany(BragiLog::class);
     }
+
+    /**
+     * List of subscription cancellations for the user
+     * @return \Illuminate\Database\Eloquent\Relations\HasMany
+     */
+    public function cancellations()
+    {
+        return $this->hasMany('App\Models\SubscriptionCancellation', 'user_id', 'id');
```
ilestis, line 176: I don't see a `where()` to limit it on cancelled subscriptions? Currently it looks for any kind of subscription, even failed or ongoing ones

resources/views/emails/subscriptions/new/html.blade.php
```diff
@@ -12,6 +12,11 @@
         Account created {{ $user->created_at->diffForHumans() }} ({{ $user->created_at->format('d.m.Y') }}).
     </p>
 
+    @if ($lastCancel)
+        <p>
+            Previously cancelled {{ $lastCancel->tier }} subscription {{ $lastCancel->created_at->diffForHumans() }} ({{ $lastCancel->created_at->format('d.m.Y') }}).
```
ilestis, line 17: I don't see a `tier` property on subscriptions. Where is it getting this from?

ilestis (changes requested):

ilestis (approved):
Ignore previous notifications, I'm hallucinating

PR #590, opened 2023-01-26
Attribute Mention Fallback

spitfire305:


app/Services/MentionsService.php
```diff
@@ -698,10 +701,18 @@ protected function mapAttributes()
         $this->text = preg_replace_callback('`\{attribute:(.*?)\}`i', function ($matches) {
             $id = (int) $matches[1];
             $attribute = $this->attribute($id);
+            $fallback = '';
+            if (str_contains($matches[1], '|')) {
+                $fallback = substr($matches[1], strpos($matches[1], "|") + 1);
```
ilestis, line 706: You can replace this by `Str::afterLast('|', $matches[1])`

app/Services/MentionsService.php
```diff
@@ -698,10 +701,18 @@ protected function mapAttributes()
...
+            }
 
-            // No entity found, the user might not be allowed to see it
+            // No entity found, the user might not be allowed to see it, if theres a fallback, apply it
             if (empty($attribute)) {
-                $replace = '<i class="unknown-mention unknown-attribute">' . __('crud.history.unknown') . '</i>';
+                if (!$fallback) {
+                    $replace = '<i class="unknown-mention unknown-attribute">' . __('crud.history.unknown') . '</i>';
+                } else {
+                    $replace = '<span class="attribute attribute-mention">' . $fallback . '</span>';
```
ilestis, line 714: Replace the classes with the 2 unknown ones so our css wizards can style links that use the fallback :)

ilestis (changes requested):
Tiny tweaks

app/Services/MentionsService.php
```diff
@@ -701,17 +701,14 @@ protected function mapAttributes()
         $this->text = preg_replace_callback('`\{attribute:(.*?)\}`i', function ($matches) {
             $id = (int) $matches[1];
             $attribute = $this->attribute($id);
-            $fallback = '';
-            if (str_contains($matches[1], '|')) {
-                $fallback = substr($matches[1], strpos($matches[1], "|") + 1);
-            }
+            $fallback = Str::after($matches[1], '|');
```
ilestis, line 704: If a mention has no fallback and can't be viewed because of permission, it now shows "attribute:123" instead of _unknown_.

ilestis (changes requested):
Not working as expected

PR #594, opened 2023-02-03
Posts css class

spitfire305:


resources/views/entities/pages/posts/_boosted.blade.php
```diff
@@ -0,0 +1,22 @@
+<table class="mb-2">
```
ilestis, line 1: I was wondering why this code was so ugly using a table, but then I saw I created this mess in the widgets 😭

ilestis (approved):

PR #599, opened 2023-02-08
Entity Logs Posts

spitfire305:


app/Http/Controllers/Entity/StoryController.php
```diff
@@ -51,6 +53,13 @@ public function save(ReorderStories $request, Entity $entity)
             ->entity($entity)
             ->reorder($request);
 
+        $log = new EntityLog();
```
ilestis, line 56: Cleaner to have this in the service, to keep controllers clean :)

app/Models/EntityLog.php
```diff
@@ -32,6 +32,10 @@ class EntityLog extends Model
     public const ACTION_UPDATE = 2;
     public const ACTION_DELETE = 3;
     public const ACTION_RESTORE = 4;
+    public const ACTION_CREATE_POST = 5;
```
ilestis, line 35: Missing new @property comments above for auto-completion in vscode/phpstorm (not related to these constants, but can't add a comment on unchanged lines)

app/Observers/PostObserver.php
```diff
@@ -71,6 +73,42 @@ public function creating(Post $post)
...
 
+    /**
+     * @param Post $post
+     */
+    public function created(Post $post)
+    {
+        $log = new EntityLog();
+        $log->entity_id = $post->entity->id;
+        $log->created_by = auth()->user()->id;
+        $log->entity_note_id = $post->id;
```
ilestis, line 84: Please replace entity_note_id with post_id, to reduce the work needed in the future, thanks :)

app/Observers/PostObserver.php
```diff
@@ -71,6 +73,42 @@ public function creating(Post $post)
...
+        // Don't log updates if just did one (typically when creating, restoring or bulk editing)
+        if (!$post->entity->hasUpdateLog() || $post->updated_at == $post->created_at || !empty($post->getOriginal('deleted_at'))) {
+            return;
+        }
+
+        $log = new EntityLog();
+        $log->entity_id = $post->entity->id;
+        $log->created_by = auth()->user()->id;
+        $log->impersonated_by = Identity::getImpersonatorId();
+        $log->entity_note_id = $post->id;
```
ilestis, line 107: Same here, `post_id` please

app/Models/EntityLog.php
```diff
@@ -77,6 +81,14 @@ public function impersonator()
         return $this->belongsTo('App\User', 'impersonated_by');
     }
 
+    /**
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function post()
+    {
+        return $this->belongsTo('App\Models\EntityNote', 'entity_note_id');
```
ilestis, line 89: Replace with `post_id` on the model, migration, and code please

app/Observers/PostObserver.php
```diff
@@ -98,6 +136,14 @@ public function saved(Post $post)
      */
     public function deleted(Post $post)
     {
+        $log = new EntityLog();
```
ilestis, line 139: I see 7 lines of code copy-pasted 3 times on this model. Please refactore to reduce the amounts of copy-pasted lines of code.

lang/en/history.php
```diff
@@ -20,6 +20,10 @@
         'delete'    => ':user deleted :entity',
         'restore'   => ':user restored :entity',
         'update'    => ':user updated :entity',
+        'create_post'    => ':user created post ":post" in :entity',
```
ilestis, line 23: `:user create the post ":post" on :entity` (missing `the` and replace `in` with `on` on all 3 keys

ilestis (changes requested):
Nice job! A few tweaks needed :)

ilestis (changes requested):
Refactor to use the same ids or adapt the filters on filtering

ilestis (changes requested):
Small refactoring and a filtering bug

app/Models/EntityLog.php
```diff
@@ -236,12 +240,15 @@ public function actions($action): array
      */
     public function scopeFilter(Builder $builder, HistoryRequest $request): Builder
     {
-        if ($request->filled('user')) {
+        if ($request->filled('user') && !$request->filled('action')) {
```
ilestis, line 243: This will become a mess if we add more filters (which we will ;-)) Please refactor so that the `orWhere` for each action is properly handled and doesn't cause copy-pasting https://stackoverflow.com/questions/51725715/orwhere-with-multiple-conditions-in-eloquent

ilestis (changes requested):
Copy-pasted code 🙈

ilestis (approved):
Wrong click sorry

PR #600, opened 2023-02-10
Recurring event: end of

spitfire305:


app/Http/Requests/AddCalendarEvent.php
```diff
@@ -31,7 +31,9 @@ public function rules()
             'year' => 'required',
             'length' => 'required|integer|min:1',
             'is_recurring' => 'nullable',
-            'recurring_until' => 'nullable',
+            'recurring_until' => 'required_with:recurring_until_month|required_with:recurring_until_day|nullable',
```
ilestis, line 34: Worth adding the `integer` rule on all 3 to avoid any shenanigans from API users.

app/Models/Calendar.php
```diff
@@ -704,6 +704,18 @@ public function upcomingReminders(int $needle = 10): Collection
                                     ->whereNull('recurring_until')
                                     // Events that end in the future are fine, they could be reoccuring on this month
                                     ->orWhere('recurring_until', '>=', $this->currentYear());
+                            })
```
ilestis, line 707: Seems to be missing from the `pastReminders()` block on the calendar dashboard widget.

app/Renderers/CalendarRenderer.php
```diff
@@ -87,6 +87,12 @@ class CalendarRenderer
      */
     protected array $eventEnd = [];
 
+    /**
+     * End date of events displayed on the calendar view
+     * @var bool
+     */
+    protected bool $eventEndsThisMonth = false;
```
ilestis, line 94: Not sure why this is a variable on the object when it's only used in a single function.

app/Renderers/CalendarRenderer.php
```diff
@@ -872,6 +879,12 @@ protected function parseReminders(Collection $reminders): void
                 if ($event->year > $this->getYear()) {
                     continue;
                 }
+                if (!empty($event->recurring_until) && $event->recurring_until == $this->getYear() && !empty($event->recurring_until_month) && $event->recurring_until_month <= $this->getMonth()) {
```
ilestis, line 882: Please add a comment explaining what this if does from a "business logic" :D

app/Renderers/CalendarRenderer.php
```diff
@@ -928,6 +941,9 @@ protected function addMultidayEvent(EntityEvent $reminder, string $date)
             $extraDate = $this->addDay($extraDate);
 
             list($y, $m, $d) = $this->splitDate($extraDate);
+            if ($this->eventEndsThisMonth == true && $d > $reminder->recurring_until_day) {
```
ilestis, line 944: Since eventEndsThisMonth is a boolean variable, no need for `== true`

database/migrations/2023_02_09_152218_update_entity_events_add_recurring_until_month_recurring_until_day.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('entity_events', function (Blueprint $table) {
+            $table->unsignedMediumInteger('recurring_until_month')->nullable();
```
ilestis, line 17: Since these fields are used in a `where()` query, we need to index them to avoid horrible performance on prod. Each should have its own index.

ilestis (changes requested):
Small tweaks to improve performance and code readability.

ilestis:
Doesn't seem to work on the yearly view. I have an event set to end in May 2023. On the yearly calendar, it's still showing in december 2023 
image
image

On the calendar dashboard, it's not seeing upcoming events happening in a few days
image
image

app/Models/Calendar.php
```diff
@@ -812,6 +823,18 @@ public function pastReminders(int $needle = 10): Collection
             ->take($needle)
             ->get();
 
+        foreach ($reminders as $key => $reminder) {
+            if (
+                (!empty($reminder->recurring_until) && empty($reminder->recurring_until_month) && $reminder->recurring_until < $this->currentYear()) ||
+                (!empty($reminder->recurring_until_month) && empty($reminder->recurring_until_day) && $reminder->recurring_until_month < $this->currentMonth() && $reminder->recurring_until < $this->currentYear()) ||
+                (!empty($reminder->recurring_until_day) && ($reminder->recurring_until_day < $this->currentDay() && $reminder->recurring_until_month == $this->currentMonth() && $reminder->recurring_until == $this->currentYear() ||
+                $reminder->recurring_until_month < $this->currentMonth() && $reminder->recurring_until <= $this->currentYear())
```
ilestis, line 831: Note for later: I'm confused by the logic here, since it's doing the same logic here compared to in the "upcoming" events loop. Shouldn't it do the reverse? And if it's the same, we should remove the copy-pasting and make it a function

app/Models/EntityEvent.php
```diff
@@ -403,12 +407,11 @@ public function mostRecentOccurrence(int $year, int $month, int $day, array $mon
         $reminderYear = $this->year;
         $reminderMonth = $this->month;
         $reminderDay = $this->day;
-
+        $overDate = false;
         // Recurring? We need to switch around the data a bit to figure out the most recent date
         if (!empty($this->is_recurring)) {
             if ($this->recurringMonthly()) {
                 //dump('monthly');
-                $reminderMonth = $month;
```
ilestis, line 411: Why remove this line? If it's recurring monthly but doesn't have more days, it will be the wrong value, no?

app/Models/EntityEvent.php
```diff
@@ -478,7 +484,9 @@ public function mostRecentOccurrence(int $year, int $month, int $day, array $mon
         if ($days > 1 && $this->length > 1) {
             $days -= $this->length - 1;
         }
-
+        if ($overDate) {
+            $days = -2;
```
ilestis, line 488: Why are we doing -2 here? It should be detailed in the code what's happening here, and why two.

app/Models/EntityEvent.php
```diff
@@ -576,6 +588,10 @@ public function nextUpcomingOccurrence(int $calendarYear, int $calendarMonth, in
         }
 
         $days += ($reminderDay - $day);
+
+        if ($overDate) {
```
ilestis, line 592: Same thing here, why -2 specifically?

app/Renderers/CalendarRenderer.php
```diff
@@ -928,6 +940,14 @@ protected function addMultidayEvent(EntityEvent $reminder, string $date)
             $extraDate = $this->addDay($extraDate);
 
             list($y, $m, $d) = $this->splitDate($extraDate);
+            //Does the event end this month? if so what day?
+            //dump($extraDate, $reminder->recurring_until_month);
+
+            if (!empty($reminder->recurring_until) && $reminder->recurring_until == $y && !empty($reminder->recurring_until_month) && $reminder->recurring_until_month <= $m) {
```
ilestis, line 946: Copy-pasted code, should be a single function if

resources/views/dashboard/widgets/calendar/body.blade.php
```diff
@@ -63,23 +63,25 @@
...
-                            @if (!empty($reminder->comment))
-                                <i class="fa-solid fa-comment" title="{{ $reminder->comment }}" data-toggle="tooltip" data-placement="bottom"></i>
-                            @endif
-                                @if ($reminder->is_recurring)
-                                <i class="fa-solid fa-arrows-rotate" title="{{ __('calendars.fields.is_recurring') }}" data-toggle="tooltip"></i>
-                            @endif
-                            <i class="fa-solid fa-calendar" title="{{ $reminder->readableDate() }}" data-toggle="tooltip" data-placement="bottom"></i>
-                        </div>
-                        {{ link_to($reminder->entity->url(), $reminder->entity->name) }}
+                    @if ($reminder->daysAgo() > -1)
```
ilestis, line 66: Nested ifs make the code more difficult to read. Instead of
```php
@if ($reminder->daysAgo() <= 0 || !$reminder->entity)
  @continue
@endif
```

resources/views/dashboard/widgets/calendar/body.blade.php
```diff
@@ -95,26 +97,28 @@
...
-                                <i class="fa-solid fa-comment" title="{{ $reminder->comment }}" data-toggle="tooltip" data-placement="bottom"></i>
-                            @endif
-                            @if ($reminder->is_recurring)
-                                <i class="fa-solid fa-arrows-rotate" title="{{ __('calendars.fields.is_recurring') }}" data-toggle="tooltip"></i>
-                            @endif
-                            @if ($reminder->isToday($calendar))
-                                <i class="fa-solid fa-calendar-check" data-toggle="tooltip" title="{{ __('calendars.actions.today') }}"></i>
-                            @else
-                                <i class="fa-solid fa-calendar" title="{{ $reminder->readableDate() }}" data-toggle="tooltip" data-placement="bottom"></i>
+                    @if ($reminder->inDays() > -1)
```
ilestis, line 100: Same story here, both ifs should be merged into an early continue statement

ilestis (changes requested):
Writing down this review for after the family tree tickets

ilestis:
Closing this as no one has requested it again

PR #602, opened 2023-02-14
Map Layer: Updating moves position bug fix

spitfire305:


resources/views/maps/groups/_form.blade.php
```diff
@@ -16,10 +16,13 @@
 
 @php 
     $options = $map->groupPositionOptions();
+    if (!empty($model->position) && $model->position == array_key_last($options) - 1) {
```
ilestis, line 19: Would be cleaner to have this in the `groupPositionOption()` function with an optional param for the current position, avoiding logic in the views as much as possible. Also a comment describing what this does would be nice because I had no idea by just reading the code :)

resources/views/maps/layers/_form.blade.php
```diff
@@ -25,6 +25,9 @@
 
 @php
     $options = $map->layerPositionOptions();
+    if (!empty($model->position) && $model->position == array_key_last($options) - 1) {
```
ilestis, line 28: Same here, move the logic to the function that generates the list, rather than adding the logic in the views :)

ilestis (changes requested):
Move logic into php instead of view

ilestis (approved):

PR #607, opened 2023-02-23
Invited user save info in settings

spitfire305:


app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -66,10 +67,16 @@ protected function validator(array $data)
      */
     protected function create(array $data)
     {
+        $settings = null;
+
+        if (Session::get('invite_token')) {
```
ilestis, line 72: We have other code that fills out the "settings" table when a user registers after clicking on a google ad. Please move the logic to the same place to keep the code clean :)

ilestis (changes requested):

PR #609, opened 2023-03-03
Admin Remove Image

spitfire305:


app/Jobs/Copyright/DeleteEntityImage.php
```diff
@@ -0,0 +1,87 @@
...
+            // Entity was deleted
+            return;
+        }
+        $service = app()->make(\App\Services\CampaignService::class);
+        $campaign = Campaign::find($entity->campaign_id);
+
+        $service->removedImage($campaign, $entity);
+        $child = $entity->child;
+
+        if (!empty($entity->child->image)) {
```
ilestis, line 71: The logic needs to be the other way round: gallery images are a superboosted campaign feature. If the campaign is superboosted, we need to remove the gallery images first, so that if the user unboosts their campaign, it falls back to the "uploaded" image in case they have uploaded one.

With the current way, if the campaign used to be superboosted and had an image, it would potentially delete a gallery image that could be attached to many other entities. To avoid that, we can only delete the gallery images if the campaign is superboosted `if ($campaign->superboosted())`

ilestis (changes requested):
Logic needs to be tweaks to avoid deleting unwanted images

PR #610, opened 2023-03-03
Quick link: tags filter not loading

spitfire305:


resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -55,7 +55,11 @@
                                 if (!empty($value) && $field['type'] == 'select2') {
                                     $modelclass = new $field['model'];
                                     $model = $modelclass->find($value);
-                                }?>
+                                }
```
ilestis, line 58: You can merge this if with the exact same if just below ;-)
```@include('...', ['value' => $filterService->filterValue('tags')])```

ilestis (changes requested):

ilestis (approved):

PR #617, opened 2023-03-15
Fixed campaign member table offset

spitfire305:


resources/views/campaigns/members/_users.blade.php
```diff
@@ -42,11 +42,11 @@
             <table id="campaign-members" class="table table-hover table-striped">
                 <thead>
                     <tr>
+                        <th></th>
                         <th>{{ __('campaigns.members.fields.name') }}</th>
                         <th>{{ __('campaigns.members.fields.roles') }}</th>
                         <th class="hidden-xs hidden-md">{{ __('campaigns.members.fields.joined') }}</th>
                         <th class="hidden-xs hidden-md">{{ __('campaigns.members.fields.last_login') }}</th>
-                        <th></th>
```
ilestis, line 49: Now the "actions" don't have a header, breaking the layout in firefox

ilestis (changes requested):

PR #624, opened 2023-03-28
Quest filter subquests

spitfire305:


app/Models/Quest.php
```diff
@@ -203,7 +206,39 @@ public function shortDescription()
...
+     */
+    public function setQuestIdAttribute($value)
+    {
+        $this->setParentIdAttribute($value);
+    }
+
+    /**
+     * Get all quests in the quests and descendants
+     */
+    public function allQuests()
```
ilestis, line 224: Don't see where this is being used, do we need it?

app/Http/Controllers/CrudController.php
```diff
@@ -550,7 +550,17 @@ protected function menuView($model, $view, $directView = false)
         $datagridSorter = $this->datagridSorter;
 
         $rows = $this->rows;
-
+        if ($show) {
```
ilestis, line 553: Remove this as it's mixing the concept of a subview with the overview

resources/views/quests/panels/quests.blade.php
```diff
@@ -13,6 +16,15 @@
             <a href="#" class="btn btn-box-tool" data-toggle="dialog" data-target="help-modal">
                 <i class="fa-solid fa-question-circle" aria-hidden="true"></i> {{ __('crud.actions.help') }}
             </a>
+            @if (request()->has('parent_id'))
+                <a href="{{ route('quests.quests', [$model]) }}" class="btn btn-box-tool">
```
ilestis, line 20: Just link to the show of the quest instead

ilestis (changes requested):

ilestis (approved):

PR #629, opened 2023-03-31
Mentions: Move to subpage.

spitfire305:


ilestis (changes requested):
Mentions to posts, the campaign, quest elements or timeline elements are no longer visible the way they were with the links to them.

app/Models/MiscModel.php
```diff
@@ -345,6 +345,17 @@ public function menuItems(array $items = []): array
             ];
         }
 
+        // Permissions for the admin?
+        if (auth()->check() && $this->entity->mentionsCount() > 0) {
+            $items['fourth']['mentions'] = [
+                'name' => 'crud.tabs.mentions',
+                'route' => 'entities.mentions',
+                'entity' => true,
+                'count' => $this->entity->mentionsCount(),
```
ilestis, line 354: I might be wrong but this looks like it's doing a complex query to the database twice on every page load, so the count needs to be cached.

app/Renderers/Layouts/Mention/Mention.php
```diff
@@ -0,0 +1,33 @@
...
+            'name' => [
+                'key' => 'name',
+                'label' => 'entities/mentions.fields.element',
+                'render' => Standard::ENTITYLINK,
+            ],
+            'type' => [
+                'key' => 'type',
+                'label' => 'crud.fields.type',
+                'render' => function ($model) {
+                    return __('entities.' . $model->entity->type());
```
ilestis, line 26: Crashed when a mention is to a campaign, and wrong when it's linked to a quest elements, timeline element, or post

ilestis (changes requested):

resources/views/layouts/datagrid/rows/entitylink.blade.php
```diff
@@ -6,6 +6,14 @@
     <?php return ?>
 @endif
 
+@if ($model instanceof \App\Models\EntityMention)
+    @if ($model->isEntity() && $model->entity->is_private || $model->isQuestElement() && $model->questElement->quest->entity->is_private || $model->isTimelineElement() && $model->timelineElement->timeline->entity->is_private || $model->isPost() && $model->post->entity->is_private)
```
ilestis, line 10: This crashes for non-admins with a mention to a private timeline or quest but where the element is visible to all. It needs all the checks on each element that "timeline" and "entity" is visible, or "quest" and "entity" is visible. It makes more sense to remove this from entitylink.blade.php and do a custom renderer in the datagrid, to avoid making the code more complicated.

app/Models/EntityMention.php
```diff
@@ -236,4 +236,63 @@ public function scopeDatagridElements(Builder $query, array $options): Builder
...
+    }
+
+    /**
+     * Get the entity link with ajax tooltip.
+     * When coming from an entity first, call this method on the entity. It avoids some back and worth.
+     * @return string
+     */
+    public function mentionLink(): string
+    {
+        if ($this->isQuestElement()) {
```
ilestis, line 264: Missing permission checks for non-admins. Page will 500 "entity is null" on most of these if permissions are set to now give visibility I believe. See how it was done in the old code where `->entity` has to be checked all the time.

ilestis (changes requested):
Crashes for non-admins, missing permission checks :)

app/Renderers/Layouts/Mention/Mention.php
```diff
@@ -17,13 +17,24 @@ public function columns(): array
             'name' => [
                 'key' => 'name',
                 'label' => 'entities/mentions.fields.element',
-                'render' => Standard::ENTITYLINK,
+                'render' => function ($model) {
+                    $row = '<i class="fa-solid fa-lock" title="' . __('crud.is_private') . '" data-toggle="tooltip"></i> ';
+
+                    if ($model->entity && !$model->entity->is_private || 
+                        $model->questElement && $model->questElement->quest && !$model->questElement->quest->entity->is_private ||
```
ilestis, line 24: Missing permission check on ->entity, also you need to add `()` around the logical checks or it won't do what you expect ;-)

app/Renderers/Layouts/Mention/Mention.php
```diff
@@ -17,13 +17,24 @@ public function columns(): array
             'name' => [
                 'key' => 'name',
                 'label' => 'entities/mentions.fields.element',
-                'render' => Standard::ENTITYLINK,
+                'render' => function ($model) {
+                    $row = '<i class="fa-solid fa-lock" title="' . __('crud.is_private') . '" data-toggle="tooltip"></i> ';
+
+                    if ($model->entity && !$model->entity->is_private || 
+                        $model->questElement && $model->questElement->quest && !$model->questElement->quest->entity->is_private || 
+                        $model->timelineElement && $model->timelineElement->timeline && !$model->timelineElement->timeline->entity->is_private ||
```
ilestis, line 25: Missing check on timeline->entity

ilestis (changes requested):
Missing one last permission check on the general entity object

ilestis (approved):

PR #633, opened 2023-04-19
Mentions: Reference values of the entity.

spitfire305:


app/Services/MentionsService.php
```diff
@@ -398,6 +398,9 @@ protected function extractAndReplace()
                     if ($field == 'race' && !$entity->child->races->isEmpty()) {
                         $data['text'] = $entity->child->races()->reorder('name')->first()->name;
                     }
+                    if ($field == 'calendar_date' && $entity->child->calendar_id) {
+                        $data['text'] = $entity->child->calendarReminder()->readableDate();
```
ilestis, line 402: Try viewing such a mention as a non-admin where the calendar is private or hidden ;-)

ilestis (changes requested):
Crash on private calendar

PR #636, opened 2023-04-21
Bug: Reorder Post Issue Fix

spitfire305:


app/Models/Entity.php
```diff
@@ -441,18 +441,15 @@ public function hasPins(): bool
...
+        return (bool) ($this->pinnedFiles->isNotEmpty());
     }
 
     /**
      * @return array|string[]
      */
     public function postPositionOptions($position = null): array
     {
-        $options = [1 => __('maps/layers.placeholders.position')];
+        $options = [null => __('characters.placeholders.dont_change'), 1 => __('maps/layers.placeholders.position')];
```
ilestis, line 452: This translation isn't related to `characters`, it would be better placed in `posts.php` along with the other translations (confused why the other fields are from map layers too haha)

ilestis (changes requested):
Translation in wrong file

app/Models/Entity.php
```diff
@@ -449,7 +449,7 @@ public function hasPins(): bool
      */
     public function postPositionOptions($position = null): array
     {
-        $options = [null => __('characters.placeholders.dont_change'), 1 => __('maps/layers.placeholders.position')];
+        $options = [null => __('entities.dont_change'), 1 => __('maps/layers.placeholders.position')];
```
ilestis, line 452: Please place it in `posts.php` where we already collect everything related to posts, and not in `entities.php` which is only related to the various entity types :)

ilestis (changes requested):
Still wrong translation file

ilestis (approved):

PR #637, opened 2023-04-24
Banned Login Log

spitfire305:


app/Http/Middleware/CheckIfUserBanned.php
```diff
@@ -21,6 +22,9 @@ public function handle(Request $request, Closure $next)
             return $next($request);
         }
 
+        $userLogType = session()->get('kanka.userLog', UserLog::TYPE_BANNED_LOGIN);
```
ilestis, line 25: I don't think this is correct: this will trigger a log when a banned user refreshes the page for the first being after being banned, which isn't useful information. Instead it could be a new key `TYPE_BANNED_LOGOUT`

ilestis (changes requested):

PR #638, opened 2023-04-26
Family Members: Add Members

spitfire305:


app/Http/Controllers/CharacterFamilyController.php
```diff
@@ -0,0 +1,58 @@
...
+        return view($this->view . '.create', [
+            'model' => $family,
+            'ajax' => $ajax
+        ]);
+    }
+
+    /**
+     * Store a newly created resource in storage.
+     */
+    public function store(Request $request, Family $family)
```
ilestis, line 44: Missing a `Request` validation that the data sent exists in the db first with user error support

app/Http/Controllers/CharacterFamilyController.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->authorize('update', $family);
+        $newMembers = 0;
+
+        foreach ($request->members as $member) {
+            if (!CharacterFamily::where('character_id', $member)->where('family_id', $family->id)->first()) {
```
ilestis, line 50: Can't we simplify this with `$family->members()->attach()` and have Laravel figure out duplicated? :)

app/Http/Controllers/CharacterFamilyController.php
```diff
@@ -0,0 +1,58 @@
...
+                $relation = CharacterFamily::create(['family_id' => $family->id, 'character_id' => $member]);
```
ilestis, line 51: Variable nether used so not needed

resources/views/families/members/_form.blade.php
```diff
@@ -0,0 +1,12 @@
+<?php
+    $selectedOption = [];
+?>
+<div class="form-group">
+    <label>{{ __('organisations.fields.members') }}</label>
+
+    <select multiple="multiple" name="members[]" id="members" class="form-control form-members" style="width: 100%" data-url="{{ route('characters.find', ['with_family' => '1']) }}">
+        @foreach ($selectedOption as $key => $val)
```
ilestis, line 8: Not sure what $selectedOption is or where it's from. Seems like this can be skipped?

routes/campaign.php
```diff
@@ -268,6 +268,7 @@
     'locations' => 'LocationController',
     //'locations.map_points' => 'LocationMapPointController',
     'families' => 'FamilyController',
+    'families.character_family' => 'CharacterFamilyController',
```
ilestis, line 271: The danger of doing it in `Resources` is that laravel is adding `index`, `show`, `get`,  `create`, `store`,  `edit`, 'update` and `delete` routes automatically. We only need two routes, so just add those two individually instead, thanks

ilestis (changes requested):

ilestis (changes requested):

PR #640, opened 2023-04-26
Race Members: Add Members

spitfire305:


resources/views/races/members/_form.blade.php
```diff
@@ -0,0 +1,4 @@
+<div class="form-group">
+    <label>{{ __('races.fields.members') }}</label>
+    <select multiple="multiple" name="members[]" id="members" class="form-control form-members" style="width: 100%" data-url="{{ route('characters.find', ['with_race' => '1']) }}"></select>
```
ilestis, line 3: What's `with_race`? I don't see anything in the code else regarding this property. Is it needed?

ilestis (changes requested):
Just one detail

ilestis (approved):

PR #642, opened 2023-04-27
API: Expose entity type IDs

spitfire305:


app/Http/Controllers/Api/v1/EntityTypeApiController.php
```diff
@@ -0,0 +1,18 @@
...
+
+class EntitytypeApiController extends ApiController
+{
+    /**
+     * @return \Illuminate\Http\Resources\Json\AnonymousResourceCollection
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function index()
+    {
+        return Resource::collection(EntityType::all());
```
ilestis, line 16: Please paginate the results in case we for one stupid reason end up with > 45 types on day :-)

ilestis (changes requested):

ilestis (approved):

PR #645, opened 2023-05-04
Users can now subscribe with paypal

spitfire305:


app/Http/Controllers/PayPalController.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+        
+        $provider = new PayPalClient;
+        $provider->setApiCredentials(config('paypal'));
+        $paypalToken = $provider->getAccessToken();
+        
+        $response = $provider->createOrder([
+            "intent" => "CAPTURE",
+            "application_context" => [
+                "return_url" => route('success.transaction'),
```
ilestis, line 39: The route name is very cryptic, would be nice to link it to Paypal :)

app/Http/Controllers/PayPalController.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * process transaction.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function processTransaction(Request $request)
+    {
+        $user = $request->user();
+        $pledge = $request->get('tier');
+        $period = $request->get('period');
```
ilestis, line 19: Doesn't seem to be used

app/Http/Controllers/PayPalController.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * success transaction.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function successTransaction(Request $request)
+    {
+        $provider = new PayPalClient;
```
ilestis, line 76: Would be nice to move all of this into a service instead of in the Controller, so we can re-use it somewhere else if we need to :)

app/Http/Requests/Settings/UserSubscribePayPal.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace App\Http\Requests\Settings;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+class UserSubscribePayPal extends FormRequest
```
ilestis, line 7: Never used?

config/paypal.php
```diff
@@ -0,0 +1,25 @@
...
+        'app_id'            => 'APP-80W284485P519543T',
+    ],
+    'live' => [
+        'client_id'         => env('PAYPAL_LIVE_CLIENT_ID', ''),
+        'client_secret'     => env('PAYPAL_LIVE_CLIENT_SECRET', ''),
+        'app_id'            => env('PAYPAL_LIVE_APP_ID', ''),
+    ],
+
+    'payment_action' => env('PAYPAL_PAYMENT_ACTION', 'Sale'), // Can only be 'Sale', 'Authorization' or 'Order'
+    'currency'       => env('PAYPAL_CURRENCY', 'USD'),
```
ilestis, line 21: Did you try doing a EUR payment? Curious if it is unhappy with this setting :)

config/paypal.php
```diff
@@ -0,0 +1,25 @@
...
+    'notify_url'     => env('PAYPAL_NOTIFY_URL', ''), // Change this accordingly for your application.
```
ilestis, line 22: What do we need to put in here? Can you put an example in the file? Thanks

routes/profile.php
```diff
@@ -161,6 +161,13 @@
 
 /*
 --------------------------------------------------------------------------
-Google2FA
+PayPal API
 --------------------------------------------------------------------------
 */
+
+Route::post('/process-transaction', 'PayPalController@processTransaction')
```
ilestis, line 168: Please prefix all these urls with `paypal/` to make it easier to see reading the code where it's going :)

ilestis (changes requested):
Awesome work! Some code quality changes

app/Services/PayPalService.php
```diff
@@ -0,0 +1,78 @@
...
+use Carbon\Carbon;
+use Srmklive\PayPal\Services\PayPal as PayPalClient;
+
+class PayPalService
+{   
+    /**
+    * @param Request $request
+    * @return mixed
+    */
+   public function process(Request $request)
```
ilestis, line 15: If I change in the HTML the pledge to "banana" before submitting the form, it crashes and will send me an email on prod, which I'd like to avoid by having some validation :)

app/Services/PayPalService.php
```diff
@@ -0,0 +1,78 @@
...
+   {
+        $user = $request->user();
```
ilestis, line 17: We also want to confirm that the user isn't already subbed, just to avoid some very angry customers and confused database (in case they open the page in two tabs)

app/Http/Controllers/PayPalController.php
```diff
@@ -1,54 +1,32 @@
...
+    /** @var PaypalService */
+    protected $service;
+
+    /**
+     * AbilityController constructor.
+     * @param PayPalService $service
+     */
+    public function __construct(PayPalService $service)
+    {
+        $this->service = $service;
```
ilestis, line 19: Missing the middlewares to check that the user is logged in and not impersonating someone else

ilestis (changes requested):
Small changes to avoid polluted data in the database for my own sanity

ilestis (approved):

PR #647, opened 2023-05-08
New posts cant save

spitfire305:


resources/views/entities/pages/posts/_form.blade.php
```diff
@@ -62,7 +62,7 @@
                 <div class="col-md-6 col-lg-4">
                     <div class="form-group">
                         <label>{{ __('crud.fields.position') }}</label>
-                        {!! Form::select('position', $options, (!empty($model->position) ? $model->position : $last), ['class' => 'form-control']) !!}
+                        {!! Form::select('position', $options, (!empty($model->position) ? 0 : $last), ['class' => 'form-control']) !!}
```
ilestis, line 65: If the position isn't empty, you're setting it to 0? That seems wrong to me

app/Models/Entity.php
```diff
@@ -450,6 +450,10 @@ public function hasPins(): bool
     public function postPositionOptions($position = null): array
     {
         $options = [null => __('posts.position.dont_change'), 1 => __('posts.position.first')];
+
+        if (!$position) {
+            $options = [1 => __('posts.position.first')];
```
ilestis, line 455: You can avoid copy-pasted content by using php's `unset($options[0])` here ;-)

ilestis (changes requested):
Not sure this works as it should

resources/views/entities/pages/posts/_form.blade.php
ilestis: Ignore me, haven't had coffee

ilestis (approved):

PR #649, opened 2023-05-09
Billing: Add info

spitfire305:


lang/en/settings.php
```diff
@@ -121,6 +121,9 @@
         'actions'   => [
             'update_profile'    => 'Update profile',
         ],
+        'billing_info' => 'Billing Information',
```
ilestis, line 124: Refactor to 
```php
'billing' => [
   'title' => 'Billing Information',
   'placeholder' => '....',
   'save' => 'Save billing information',
],
```

So that it has the same "system" as the rest of kanka where things that belong together are in their own array :)

app/Observers/UserObserver.php
```diff
@@ -43,6 +43,19 @@ public function saving(User $user)
             }
         }
 
+        //Purify Billing info
+        if (!empty($user->profile['billing'])) {
+            $profile = $user->profile;
+            $profile['billing'] = mb_substr(strip_tags($profile['billing']), 0, 1024);
```
ilestis, line 49: Replace strip_tags with the "purify" system we use to avoid people writing javascript and other nonsense

resources/views/vendor/cashier/receipt.blade.php
```diff
@@ -0,0 +1,323 @@
...
+
+                @if ($invoice->customer_name)
+                    {{ $invoice->customer_email }}<br>
+                @endif
+
+                @foreach ($invoice->customerTaxIds() as $taxId)
+                    {{ $taxId->value }}<br>
+                @endforeach
+                
+                {{ $billing }}
```
ilestis, line 157: I have a suspicion that strange things happen when someone writes an address with an "&", an "#", an "é" or other special characters ;-) Does multi-line info work?

app/Models/UserSetting.php
```diff
@@ -224,6 +224,20 @@ public function resetTutorials(): self
...
 
+    /**
+     * Save user's custom billing info
+     * @return $this
+     */
+    public function updateBillingInfo($billing): self
+    {
+        $profile = $this->profile;
+        $profile['billing'] = $billing;
+
```
ilestis, line 235: Missing an unset if the user is removing all of their billing info

app/Http/Controllers/Settings/ProfileController.php
```diff
@@ -46,7 +46,23 @@ public function update(StoreSettingsProfile $request)
             ->with('success', trans('settings.profile.success'));
     }
 
-        /**
+    /**
+     * @param StoreSettingsProfile $request
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function saveBillingInfo(Request $request)
```
ilestis, line 53: The request should throw an error if the info is longer than 1024 characters, rather than just delete it without the user being warned that we removed content

ilestis (changes requested):

ilestis (approved):

PR #652, opened 2023-05-15
Database Population.

spitfire305:


app/Console/Commands/DummyEntities.php
```diff
@@ -0,0 +1,180 @@
...
+     * @return int
+     */
+    public function handle()
+    {
+        $campaignId = $this->argument('campaign');
+        $this->loadObservers();
+
+        //Generate Characters and Locations
+        Location::factory()
+            ->state(['campaign_id' => $campaignId, 'name' => 'Thaelia', 'slug' => Str::slug('Thaelia')])
```
ilestis, line 68: You no longer need to set the slug as that is now handled in the observers :-)
I also think you can skip the `campaign_id` set now that you have the observers, but could be wrong

ilestis (changes requested):
Nice work! I think the code can be simplified :)

ilestis (approved):

PR #653, opened 2023-05-16
Entity Transform: Locations to location

spitfire305:


app/Services/EntityService.php
```diff
@@ -758,7 +781,9 @@ protected function fixTree(MiscModel $model): void
         if (!method_exists($model, 'recalculateTreeBounds')) {
             return;
         }
-        $model->setParentId(null);
+        if (!in_array('parent_location_id', $model->getFillable())) {
```
ilestis, line 784: I'm confused by this check, how do we fix location trees now as they have a parent_location_id? Why is this if needed?

ilestis (changes requested):

ilestis (approved):
🥳

PR #656, opened 2023-05-19
Users can now set focus points for gallery images

spitfire305:


app/Http/Controllers/Campaign/GalleryController.php
```diff
@@ -125,6 +126,34 @@ public function edit(Image $image)
...
+     * @param Entity $entity
+     * @return \Illuminate\Http\RedirectResponse
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function saveFocus(StoreImageFocus $request, Image $image)
+    {
+        $campaign = CampaignLocalization::getCampaign();
+        $this->authorize('gallery', $campaign);
+
+        $image->focus_x = $request->post('focus_x');
```
ilestis, line 140: Please move the code into the service instead of the controller so that we can re-use the code elsewhere if needed :-)

database/migrations/2023_05_17_191929_update_images_add_focus_x_focus_y.php
```diff
@@ -0,0 +1,30 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('images', function (Blueprint $table) {
+            $table->integer('focus_x')->nullable();
```
ilestis, line 15: Please make it unsigned since we don't need negative positions

resources/views/gallery/_image.blade.php
```diff
@@ -24,9 +24,15 @@
                 <x-icon class="fa-regular fa-file text-4xl"></x-icon>
             </div>
         @else
-        <a class="block avatar grow relative cover-background"
-             style="background-image: url('{{ Img::crop(192, 144)->url($image->path) }}')">
-        </a>
+            @if ($image->focus_x && $image->focus_y)
```
ilestis, line 27: This should be a function on the Model that takes care of providing the image url with the focus or not

resources/views/gallery/edit.blade.php
```diff
@@ -104,6 +110,32 @@
...
+        <div class="">
+            <x-box>
+                <p class="help-block">{{ __('entities/image.focus.helper') }}</p>
+                <div class="focus-selector max-h-screen relative mb-2 overflow-auto">
+                    <div class="focus absolute text-white cursor-pointer text-3xl" style="@if(empty($image->focus_x))display: none; @else left: {{ $image->focus_x }}px; top: {{ $image->focus_y }}px; @endif">
+                        <i class="fa-regular fa-bullseye fa-2x hover:text-red" aria-hidden="true"></i>
+                    </div>
+                    <img class="focus-image max-w-none" src="{{ $image->getImagePath(0) }}" alt="img" />
+                </div>
+                <div class="flex gap-2 sm:gap-5 items-center mb-5 pull-right">
```
ilestis, line 124: Don't think that `pull-right` is needed here, instead the first form can have a "grow" class to make it stretch, pushing the save to the right as from the ticket's mockup

resources/js/story.js
```diff
@@ -1,18 +1,22 @@
...
+        initImageFocus()
+     });
 });
 
 function initImageFocus() {
     let target = $('.focus-image');
     if (target.length === 0) {
         return;
     }
+    console.log('its starting');
```
ilestis, line 13: Please comment the console logs :-)

ilestis (changes requested):
Nice job! A bit of refactoring, but nothing major

ilestis (changes requested):
When selecting the image from the dropdown, it isn't focused
image

When viewing an entity, the image isn't focused
image

When editing the entity, the image preview in the premium tab isn't focused
image

In the old table view, the image thumbnail isn't focused
image

In the relations explorer, the thumbnail isn't focused
image

In the list of a character's orgs
image

When editing a character's races
image

In general the dropdowns seem to never focus, probably because it uses another code that we can replace with the new avatarv2 one. Most of the issues could be fixed by adding the focus on the `MiscModel::getImageFallback()` function.

ilestis (approved):

PR #660, opened 2023-05-23
Atribute Templates: Entity List

spitfire305:


app/Models/EntityType.php
```diff
@@ -47,6 +48,11 @@ public function getClass()
      */
     public function name(): string
     {
+        $campaign = CampaignLocalization::getCampaign();
```
ilestis, line 51: This whole function can be replaced by a single call to the Module facade :-)

ilestis (changes requested):
Use the new Module facade to get the custom module name directly

ilestis (approved):

PR #661, opened 2023-05-24
Unpublished Plugins are now hidden

spitfire305:


resources/views/campaigns/plugins/_table.blade.php
```diff
@@ -18,11 +18,13 @@
         </thead>
         <tbody>
         @foreach ($rows as $row)
-            <tr class="@if (Datagrid::isHighlighted($row)) warning row-highlighted @endif">
-                @foreach (Datagrid::columns($row) as $column)
-                    @include('layouts.datagrid._column')
-                @endforeach
-            </tr>
+            @if ($row->status_id != 1)
```
ilestis, line 21: This should be in the sql query, not in the view :) Also does this work with draft plugins a user is working on? Because we have an id especially for hidden plugins (see Status.php on the marketplace)

ilestis (changes requested):
Breaks draft plugins

resources/views/campaigns/plugins/_table.blade.php
ilestis: Just realised this is to unpublish a plugin that was published by mistake, so ignore this. We don't need to make any changes in Kanka. Only fix the code in the marketplace

PR #663, opened 2023-05-30
Family trees v2

spitfire305:


app/Services/Families/FamilyTreeService.php
```diff
@@ -110,7 +113,7 @@ protected function prepareEntities(): void
             if ($k !== 'entity_id') {
                 return;
             }
-            if ($v === 0 || in_array($v, $this->configEntityIds)) {
+            if ($v === 0 || in_array($v, $this->configEntityIds) || $v === null) {
```
ilestis, line 116: Why a check on `null`? If it's needed, we can combine `0` and `null` with `empty($v)`

app/Services/Families/FamilyTreeService.php
```diff
@@ -170,6 +174,62 @@ protected function formatEntity(Entity $entity): array
...
+
+protected function visibilityCheck(): void
+{
+    $config = [];
+    foreach ($this->config as $key => $node) {
+        $config[$key] = $this->cleanInvisible($node, $key);
+    }
+    $this->config = $config;
+
+}
```
ilestis, line 186: Missing newline after the function end

resources/js/components/families/FamilyParentChildrenLine.vue
```diff
@@ -15,6 +15,7 @@ export default {
         index: 0,
         column: undefined,
         row: undefined,
+        node: '',
```
ilestis, line 18: Same here, if node is an object, `undefined` is a better default

ilestis (changes requested):
Small refactoring changes

ilestis (approved):
🥳

PR #664, opened 2023-05-30
Entity Aliases: Pinned

spitfire305:


ilestis (changes requested):
When viewing an entity as a non admin, if there is **only** a pinned alias, the "pinned" section isn't visible

ilestis (approved):

PR #671, opened 2023-06-05
Gallery: Track inserted Images.

spitfire305:


app/Models/ImageMention.php
```diff
@@ -0,0 +1,128 @@
...
+
+    protected $sortable = [
+        'name',
+        'type',
+    ];
+
+    /**
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function target()
```
ilestis, line 41: `image()` would be a nicer relation name

app/Models/ImageMention.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return $this->belongsTo('App\Models\Images', 'image_id', 'id');
```
ilestis, line 43: I don't see a model called `Images` (plural). Does this function get called? If it never gets called, we can remove it altogether.

app/Models/ImageMention.php
```diff
@@ -0,0 +1,128 @@
...
+        return !empty($this->post_id);
+    }
+
+    /**
+     * Determine if the mention goes to an entity
+     * @return bool
+     */
+    public function isEntity(): bool
+    {
+        return !empty($this->entity_id);
```
ilestis, line 77: As per the requirement of an image mention always having an entity_id, this function will always return true, and thus be useless.

app/Services/EntityMappingService.php
```diff
@@ -59,7 +61,8 @@ public function mapEntity(Entity $entity)
      */
     public function mapPost(Post $post)
     {
-        return $this->map($post);
+        //return $this->map($post);
+        return $this->images($post)->map($post);
```
ilestis, line 65: I don't see the main entity's entry field called `images()` ?

app/Models/ImageMention.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * @param Builder $query
+     * @return Builder
+     */
+    public function scopePost(Builder $query): Builder
+    {
+        return $query->whereNotNull('image_mentions.entity_note_id');
+    }
+
+    public function linkToEntity(): string
```
ilestis, line 120: Having a Model (part of the db) know about HTML is dirty. I know some older parts of Kanka have that, but we should avoid putting html in models. Please move the html part into the view, thanks :)

ilestis (changes requested):
Looks good! A few small refactoring bits, but also I'm not seeing the code being ever called for the entry field of an entity, only on posts 🤔

ilestis (approved):
Nice!

PR #673, opened 2023-06-06
Tags are now removed when transforming an entity into a tag.

spitfire305:


app/Services/EntityService.php
```diff
@@ -452,6 +452,11 @@ protected function moveType(Entity $entity, string $target, MiscModel $misc = nu
             }
         }
 
+        //Remove tags if converting to tag, since tags can have tags.
```
ilestis, line 455: I think you meant `can't` :P

app/Services/EntityService.php
```diff
@@ -452,6 +452,11 @@ protected function moveType(Entity $entity, string $target, MiscModel $misc = nu
...
+        if ($new->entityTypeId() === config('entities.ids.tag')) {
+            $old->entity->tags()->detach();
```
ilestis, line 457: Instead of calling $old->entity, which creates a new db request for just this, use `$entity` directly instead

ilestis (changes requested):

ilestis (approved):

PR #674, opened 2023-06-07
Reminder: Change entity

spitfire305:


resources/views/calendars/events/_form.blade.php
```diff
@@ -12,7 +12,9 @@
         </span>
     </x-grid>
 @else
-    @include('cruds.fields.entity_link', ['entity' => $entityEvent->entity])
+    <div class="field-entity">
```
ilestis, line 15: image
Double "field-entity" block, but it does need a `mb-5` class to add some bottom margin to make it look aligned with the rest of the form

ilestis (approved):

PR #678, opened 2023-06-12
Admin: campaign image

spitfire305:


app/Jobs/Copyright/DeleteCampaignImage.php
```diff
@@ -0,0 +1,59 @@
...
+    use SerializesModels;
+
+    protected $campaignId;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Request $request)
```
ilestis, line 29: Constructor not needed since it's sanitized and de-sanitized directly by the queue

ilestis (changes requested):

ilestis (approved):

PR #681, opened 2023-06-14
Console Command: Delete Campaign

spitfire305:


app/Console/Commands/CampaignDelete.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $campaignId = $this->argument('campaign');
+        $campaign = Campaign::where('id', $this->argument('campaign'))->first();
```
ilestis, line 44: Code crashes if I give an invalid id :)

app/Console/Commands/CampaignDelete.php
```diff
@@ -0,0 +1,50 @@
...
+
+        CampaignDeleteJob::dispatch($campaign);
+
```
ilestis, line 47: Would be nice to have a message that the campaign id was queued for deletion

ilestis (changes requested):

ilestis (approved):

PR #685, opened 2023-06-21
Features/logs v2

spitfire305:


app/Http/Kernel.php
```diff
@@ -49,6 +49,7 @@ class Kernel extends HttpKernel
         'api' => [
             //Do this in the routes 'throttle:rate_limit,1',
             'bindings',
+            'api.log',
```
ilestis, line 52: You can just write `\App\Http\Middleware\ApiLogMiddleware::class` instead of `'api.log'` and remove it later

app/Http/Kernel.php
```diff
@@ -95,6 +96,7 @@ class Kernel extends HttpKernel
         'subscriptions' => \App\Http\Middleware\Subscriptions::class,
         'fullsetup' => FullSetup::class,
         '2fa' => \App\Http\Middleware\OTPMiddleware::class,
+        'api.log' => \App\Http\Middleware\ApiLogMiddleware::class,
```
ilestis, line 99: And remove this one since we're not re-using it. I think

database/migrations/2023_06_10_220139_create_api_logs.php
```diff
@@ -0,0 +1,37 @@
...
+        if (!env('DB_LOGS_DATABASE', false)) {
+            return;
+        }
+        Schema::connection('logs')->create('api_logs', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
+            $table->unsignedInteger('user_id')->nullable();
+            $table->unsignedInteger('campaign_id')->nullable();
+            $table->text('uri')->nullable();
+            $table->json('params')->nullable();
```
ilestis, line 23: I forgot that we need an index on `created_at` since we're using it in the `WHERE` for deletion :(

database/migrations/2023_06_15_210220_create_user_logs_table.php
```diff
@@ -1,37 +1,32 @@
...
-            $table->increments('id');
+        Schema::connection('logs')->create('user_logs', function (Blueprint $table) {
+            $table->id();
             $table->integer('user_id')->unsigned();
             $table->unsignedTinyInteger('type_id')
                 ->default(\App\Models\UserLog::TYPE_LOGIN);
             $table->string('ip', 255)->nullable();
+            $table->char('country', 6)->nullable();
             $table->timestamps();
-
```
ilestis, line 23: Same here, since we're recreating the user_logs table, might as well add the index for the purging of old rows

ilestis (changes requested):
Small changes, mostly because I'm a dumb dumb

ilestis (approved):

PR #686, opened 2023-06-21
Refactor: Pinned Trait

spitfire305:


resources/api-docs/1.0/relations.md
```diff
@@ -32,6 +32,7 @@ You can get a list of all the relations of an entity by using the following endp
             "attitude": 22,
             "visibility_id": 1,
             "is_star": false,
```
ilestis, line 34: We can remove `is_star` from the docs, because we want people to use `is_pinned` instead

resources/api-docs/1.0/attributes.md
```diff
@@ -97,6 +99,7 @@ To create an attribute, use the following endpoint.
 | `entity_id` | `int` (Required) | The attribute's parent entity |
 | `is_private` | `boolean` | If the attribute is only visible to `admin` members of the campaign |
 | `is_star` | `boolean` | If the attribute is "pinned" on the entity view |
```
ilestis, line 101: We can remove `is_star` from the docs, because we want people to use `is_pinned` instead

resources/api-docs/1.0/relations.md
```diff
@@ -61,6 +62,7 @@ To get the details of a single relation, use the following endpoint.
         "attitude": 22,
         "visibility_id": 1,
         "is_star": true,
```
ilestis, line 64: Same here

resources/api-docs/1.0/relations.md
```diff
@@ -90,6 +92,7 @@ To create a relation, use the following endpoint.
 | `colour` | `string` | Hex colour of the attitude (with or without the `#`) |
 | `two_way` | `boolean` | If set, will duplicate the relation but in the other direction |
 | `is_star` | `boolean` | If the relation is visible on the entity's submenu |
```
ilestis, line 94: Same here

app/Models/Scopes/Pinnable.php
```diff
@@ -5,12 +5,12 @@
 use Illuminate\Database\Eloquent\Builder;
 
 /**
- * Trait Pinned
+ * Trait Pinnable
  * @package App\Models\Scopes
```
ilestis, line 9: We can add a `@property bool $is_pinned` here to make help with the IDE auto-complete code, and now need it on the models themselves :)

app/Models/Scopes/Pinnable.php
```diff
@@ -5,12 +5,12 @@
...
  *
- * @method static self|Builder pinned(bool $is_pinned = true)
+ * @method static self|Builder pinnable(bool $is_pinned = true)
  */
-trait Pinned
+trait Pinnable
 {
     /**
      * @param Builder $query
```
ilestis, line 16: For extra niceness, add a new public function `isPinned(): bool` and use that throughout the app, it's a bit cleaner than calling the db property directly.

ilestis (changes requested):
Nice job! Small refactoring needed to make it perfect

ilestis (approved):

PR #687, opened 2023-06-22
Summernote: Remove Summary

spitfire305:


ilestis (changes requested):
Missing removing the plugin from the summernote initialisation code itself

ilestis (approved):

PR #688, opened 2023-06-22
Duplicate Role

spitfire305:


app/Http/Controllers/CampaignRoleController.php
```diff
@@ -98,6 +121,15 @@ public function store(StoreCampaignRole $request)
         }
         $this->authorize('create', CampaignRole::class);
         $role = CampaignRole::create($request->all());
+        if ($request->has('duplicate') && $request->get('duplicate') != 0) {
```
ilestis, line 124: This should be in some service rather than in the controller :)

app/Http/Controllers/CampaignRoleController.php
```diff
@@ -98,6 +121,15 @@ public function store(StoreCampaignRole $request)
...
+            $oldRole = CampaignRole::where('id', $request->get('role_id'))->first();
+            foreach ($oldRole->permissions as $permission) {
+                /** @var CampaignPermission $newPermission */
+                $newPermission = $permission->replicate(['user_id', 'campaign_id', 'key', 'action', 'table_name', 'entity_type_id', 'access', 'entity_id', 'misc_id']);
```
ilestis, line 128: Replicate's first param is a list of configs that are ignored. I'm surprised that this works? Normally it should just be `->replicate(['campaign_role_id'])` 🤔

app/Http/Controllers/CampaignRoleController.php
```diff
@@ -98,6 +121,15 @@ public function store(StoreCampaignRole $request)
...
+                $newPermission->campaign_role_id = $role->id;
+                $newPermission->saveQuietly();
```
ilestis, line 130: We can do a normal save() since it's a new permission and we want the timestamps

app/Renderers/Layouts/Campaign/CampaignRole.php
```diff
@@ -95,6 +95,13 @@ public function actions(): array
                 'icon' => 'fa-solid fa-edit',
                 'route' => 'campaign_roles.show',
             ],
+            'duplicate' => [
+                'label' => 'campaigns.roles.actions.duplicate',
+                'icon' => 'fa-solid fa-edit',
```
ilestis, line 100: Icon can be "fa-solid fa-copy"

resources/views/campaigns/roles/_form.blade.php
```diff
@@ -2,5 +2,17 @@
 <div class="field-name required">
     <label>{{ trans('campaigns.roles.fields.name') }}</label>
     {!! Form::text('name', null, ['placeholder' => trans('campaigns.roles.placeholders.name'), 'class' => 'form-control', 'maxlength' => 45, 'required']) !!}
+    @if (isset($roleId))
+        <div class="field-skip-zero checkbox">
```
ilestis, line 6: `field-skip-zero`? :) We give each field it's own css class name to allow people to style it. Should be `field-duplicate`

ilestis (changes requested):

ilestis (approved):

PR #691, opened 2023-06-28
Campaigns: Genre

spitfire305:


app/Http/Controllers/Search/MiscController.php
```diff
@@ -329,24 +329,32 @@ public function attributeTemplates(Request $request)
      */
     protected function buildSearchResults(string $term, string $class, array $excludes = [])
     {
+        $name = 'name';
```
ilestis, line 332: I feel like this function now does too much, mixing many ideas together. Can you revert it and have a different buildGenres() function that just focuses on genres? Thanks!

app/Models/CampaignGenre.php
```diff
@@ -0,0 +1,44 @@
...
+ * @package App\Models
+ *
+ * @property int $genre_id
+ * @property int $campaign_id
+ * @property Campaign $campaign
+ * @property Genre $genre
+ */
+class CampaignGenre extends Pivot
+{
+    use Paginatable;
```
ilestis, line 19: I don't think that it needs to be paginatable

app/Http/Requests/FilterPublicCampaignRequest.php
```diff
@@ -29,7 +29,8 @@ public function rules()
             'featured_until' => 'nullable|boolean',
             'is_open' => 'nullable|boolean',
             'system' => 'nullable|string|max:25',
-            'language' => 'nullable|string|max:5'
+            'language' => 'nullable|string|max:5',
+            'genre' => 'nullable|string|max:25'
```
ilestis, line 33: This says it's a string max 25 but in the code it looks like it's testing it on the id in `scopeFilterPublic`

app/Observers/CampaignObserver.php
```diff
@@ -225,4 +227,40 @@ protected function saveRpgSystems(Campaign $campaign): void
...
+    protected function saveGenres(Campaign $campaign)
+    {
+        if (!request()->has('genres')) {
+            return;
+        }
+        $ids = request()->post('genres', []);
+
+        $existing = [];
+        /** @var Genre $genre */
+        foreach ($campaign->genres()->get() as $genre) {
```
ilestis, line 243: `->genres()->get()` can be simplified as `->genres`

app/Services/GenreService.php
```diff
@@ -0,0 +1,23 @@
...
+namespace App\Services;
+
+use App\Models\Genre;
+
+class GenreService
+{
+    /**
+     * @return array
+     */
+    public function getGenres($emptyOption = false)
```
ilestis, line 12: Missing the parameter type casting, and the function return casting

database/migrations/2023_06_27_194149_create_genres_table.php
```diff
@@ -0,0 +1,32 @@
...
+            $table->index(['slug']);
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('genres', function (Blueprint $table) {
```
ilestis, line 28: Should just be the dropIfExists, without the Schema::table

database/migrations/2023_06_27_195226_create_campaign_genre_table.php
```diff
@@ -0,0 +1,34 @@
...
+            $table->foreign('genre_id')->references('id')->on('genres')->cascadeOnDelete();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('campaign_genre', function (Blueprint $table) {
```
ilestis, line 30: Same here, down() should only be a single line :)

database/migrations/2023_06_27_194149_create_genres_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('genres', function (Blueprint $table) {
+            $table->increments('id');
+            $table->unsignedInteger('campaign_count')->default(0);
+            $table->string('slug', 50);
```
ilestis, line 17: slugs need to be unique, forgot about that detail

database/migrations/2023_06_27_194149_create_genres_table.php
```diff
@@ -0,0 +1,32 @@
...
+            $table->timestamps();
+            $table->index(['slug']);
```
ilestis, line 19: If the slug is unique the index isn't needed (it's automatic)

resources/views/front/campaigns.blade.php
```diff
@@ -61,6 +63,9 @@
                     <div class="m-1">
                         {!! Form::select('featured_until', ['' => __('front.campaigns.open.filters.featured'), 0 => __('general.yes'), 1 => __('general.no')], request()->get('featured_until'), ['class' => 'form-control']) !!}
                     </div>
+                    <div class="m-1">
+                        {!! Form::select('genre', ['' => __('campaigns.fields.genre')] + $genres->getGenres(), request()->get('genre'), ['class' => 'form-control']) !!}
```
ilestis, line 67: getGenres() has a param for an empty option, looks like it could be merge with the __('campaigns.fields.genre')

resources/views/components/form/genres.blade.php
```diff
@@ -0,0 +1,32 @@
...
+elseif(!empty($model)) {
+    /** @var \App\Models\Genre $genre */
+    foreach ($model->genres as $genre) {
+        $selectedOption[$genre->id] = __('genres.' . $genre->slug);
+    }
+}
+?>
+<label>{{ __('entities.families') }}</label>
+
+<select multiple="multiple" name="genres[]" class="form-control select2 join-item" data-tags="true" style="width: 100%" data-url="{{ route('genres.find') }}" data-allow-clear="true" data-allow-new="false" data-placeholder="" id="{{ $fieldUniqIdentifier }}">
```
ilestis, line 27: I'm not seeing where it's being limited to max 3 genres. It should also display all 20 options in the dropdown at all times because users won't be able to guess what are the available options. Instead of loading them via ajax, we can enable them directly in the page.

ilestis (changes requested):
Small tweaks and confirming that the genres field is limited to 3 results

app/Observers/CampaignObserver.php
```diff
@@ -233,14 +233,11 @@ protected function saveRpgSystems(Campaign $campaign): void
      */
     protected function saveGenres(Campaign $campaign)
     {
-        if (!request()->has('genres')) {
```
ilestis: Why remove this? Now when I update the campaign by installing a plugin, all the genres get removed :)

resources/js/components/select2.js
```diff
@@ -8,17 +8,26 @@ $(document).ready(function() {
...
+                    maximumSelectionLength: 3,
+                });
+                return;
+            }
 
             // Check it isn't the select2-icon
             let allowClear = $(this).data('allow-clear');
             let dropdownParent = $(this).data('dropdown-parent');
-
+            console.log($(this).data());
```
ilestis, line 26: Don't think we need this

resources/js/components/select2.js
```diff
@@ -8,17 +8,26 @@ $(document).ready(function() {
...
             $(this).select2({
                 tags: false,
                 placeholder: $(this).data('placeholder'),
                 allowClear: allowClear || true,
-                //tags: $(this).data('tags') || false,
-                language: $(this).data('language'),
```
ilestis, line 21: Now the dropdown is no longer translated :(

ilestis (changes requested):
Small cleanup tweaks

resources/js/components/select2.js
spitfire305: Sorry hahaha

ilestis (approved):

PR #692, opened 2023-06-28
Random Widget: Option to use header

spitfire305:


resources/views/dashboard/widgets/forms/_random.blade.php
```diff
@@ -53,6 +53,16 @@
             @includeWhen(!$boosted, 'dashboard.widgets.forms._boosted')
 
             <x-grid>
+                {!! Form::hidden('config[entity-header]', 0) !!}
```
ilestis, line 56: This smells copy-pasted, anyway we can avoid writing the same code twice? :)

ilestis (changes requested):

resources/views/dashboard/widgets/forms/_preview.blade.php
```diff
@@ -45,18 +45,8 @@
 
             <x-grid>
                 <div>
```
ilestis, line 47: Why do you keep the div around the include in this form, but not in the other form? Seems like it's useless here

ilestis (changes requested):

ilestis (approved):

PR #694, opened 2023-07-06
Non-Admin Post Editing

spitfire305:


ilestis (approved):
That's one ugly if but it seems to work

PR #695, opened 2023-07-06
User Log PII Cleanup

spitfire305:


app/Services/Users/UserLogService.php
```diff
@@ -0,0 +1,16 @@
...
+namespace App\Services\Users;
+
+use App\Models\UserLog;
+
+class UserLogService
+{
+    public function deleteOldLogs(): self
+    {
+        $logs = UserLog::where('created_at', '<=',\Carbon\Carbon::today()->subDays(30)->format('Y-m-d'));
+        $logs->update(array('ip' => null, 'country' => null));
```
ilestis, line 12: I don't think you even need to declare the variable

ilestis (changes requested):
Tiny change

PR #700, opened 2023-07-14
Post layout phase 2

spitfire305:


ilestis (changes requested):
Small changes, mostly about re-using existing translations

ilestis (approved):

PR #703, opened 2023-07-18
Refactor/api subdomain

ilestis:


ilestis:
This was handled better in the kanka v2 PR

PR #705, opened 2023-07-18
Dashboard Widgets: Show members sort

spitfire305:


resources/views/dashboard/widgets/previews/_members.blade.php
```diff
@@ -18,8 +18,10 @@
 }
 $members = $entity->isFamily()
     ? $child->members()->orderBy('name')->get()
-    : $child->members()->with(['character', 'character.entity'])->get()
-;
+    : $child->members()->with(['character', 'character.entity'])
+        ->orderBy(
+            App\Models\Character::select('name')->whereColumn('organisation_member.character_id', 'characters.id'), 'asc'
```
ilestis, line 23: This seem overly complicated, why not make a left join and order on that?

ilestis (changes requested):

ilestis (approved):

PR #707, opened 2023-07-19
Location: Module Icon

spitfire305:


resources/views/entities/components/_post.blade.php
```diff
@@ -71,7 +71,7 @@
 
                 @if ($post->location)
                 <span class="entity-note-detail-element entity-note-location post-detail-element post-location">
-                    <x-icon class="ra ra-tower"></x-icon>
+                <x-icon class="{{ \App\Facades\Module::icon(config('entities.ids.location'), 'ra ra-tower') }}"></x-icon>
```
ilestis, line 74: You can switch it to `:class="\App\Facades\...."`. If a blade component variable is prefixed with `:`, it will interpret the value as php already

ilestis (changes requested):
There is a better way :-)

ilestis (approved):

PR #711, opened 2023-07-25
Pricing Page

spitfire305:


resources/views/settings/subscription/index.blade.php
```diff
@@ -6,6 +6,17 @@
  */
 
 ?>
+@php
+    $country = 'CH';
```
ilestis, line 10: Please put this in the controller, or better yet in a service, so we can re-use it on the other subscriptions page :)

resources/views/settings/subscription/index.blade.php
```diff
@@ -58,9 +69,15 @@
             <span class="slider subscription-period-slider"></span>
         </label>
         <span>{{ __('tiers.toggle.yearly') }}</span>
+
+        <select class="custom-select custom-select-lg mb-3" id="currency-selector">
+            <option selected>Currency</option>
+            <option value="eur">EUR</option>
+            <option value="usd">USD</option>
```
ilestis, line 76: The USD option in the dropdown isn't selected by default when outside of the few countries using EUR

resources/views/settings/subscription/index.blade.php
```diff
@@ -6,6 +6,17 @@
  */
 
 ?>
+@php
```
ilestis, line 9: This should be on the _pricing_ page, not the _subscription_ page, where we already have a currency changer

ilestis (changes requested):
Wrong page :D

PR #715, opened 2023-07-27
Merge Tags

spitfire305:


app/Http/Controllers/TagController.php
```diff
@@ -181,4 +189,51 @@ public function entityStore(StoreTagEntity $request, Tag $tag)
...
+    public function transferTag(Tag $tag)
+    {
+        $this->authorize('update', $tag);
+
+        $models = Tag::where('campaign_id', $tag->campaign_id)->get();
+
+        $tags[''] = __('tags.transfer.placeholder');
+
+        foreach ($models as $model) {
+            if ($tag->id != $model->id) {
```
ilestis, line 207: Several campaigns have over 1000 tags. We need to use the ajax dropdown, otherwise we'll crash the server and kill our bandwidth.

ilestis (changes requested):

resources/views/tags/transfer.blade.php
```diff
@@ -17,12 +17,17 @@
             <p class="help-block mb-5">
                 {{ __('tags.transfer.description') }}
             </p>
-
             <div class="field-campaign mb-5">
                 <label>{{ __('tags.fields.tag') }}</label>
-                {!! Form::select('tag', $tags, null, ['class' => 'form-control']) !!}
+                <select  name="tag"
```
ilestis, line 22: You can replace the label and field (because right now it's also not using the custom module name for tags) with only `@include('cruds.fields.tag', ['model' => $tag, 'allowNew' => false])` :-)  And replace the field name from `tag` to `tag_id` to keep the code consistant (change in service and request validator)

ilestis (changes requested):
Nearly!

lang/en/tags.php
```diff
@@ -17,6 +17,7 @@
         'children'          => 'Children',
         'is_auto_applied'   => 'Automatically apply to new entities',
         'is_hidden'         => 'Hidden from header and tooltip',
+        'tag'               => 'Tag',
```
ilestis, line 20: Can be deleted :-)

lang/en/tags.php
```diff
@@ -36,4 +37,12 @@
             'children'  => 'Children',
         ],
     ],
+    'transfer' => [
+        'title'         => 'Transfer :name',
+        'description'   => 'Move this tag\'s entities to another tag.',
+        'transfer'      => 'Transfer',
+        'success'       => 'Succesfully transfered entities from :tag to :newTag',
+        'fail'          => 'Failed to transfer entities from :tag to :newTag',
+        'placeholder'   => 'Select a tag',
```
ilestis, line 46: Can also be deleted

ilestis (changes requested):
Cleanup of old code

PR #719, opened 2023-07-31
Campaign Invited: Deletion confirmation

spitfire305:


resources/views/campaigns/members/_invites.blade.php
```diff
@@ -109,4 +108,29 @@
             <div class="modal-content rounded-2xl bg-base-100" id="small-modal-content"></div>
         </div>
     </div>
+
+    <div class="modal fade" id="datagrid-bulk-delete" tabindex="-1" role="dialog" aria-labelledby="clickConfirmLabel">
```
ilestis, line 112: We already have a modal for deleting confirmation embedded in every single page. No need to create a new one.

ilestis (changes requested):
Page crashes when I try and open it and the campaign has 0 invites.
When I have multiple invite links, it uses the text from the last one.

ilestis (approved):

PR #721, opened 2023-08-01
Reminder: Too long

spitfire305:


app/Services/LengthValidatorService.php
```diff
@@ -0,0 +1,40 @@
...
+        $monthLength = 0;
+        foreach ($calendar->monthDataProperties() as $monthData) {
+            $counter = $counter + 1;
+            if ($counter >= $month) {
+                $monthLength = $monthLength + $monthData['data-length'];
+            }
+        }
+        $totalLength = $monthLength - $day + $daysInYear;
+        if ($length >= $totalLength) {
+            return json_encode([
```
ilestis, line 30: Shouldn't be the service's job to return an encoded json, but the controller :-)

app/Http/Controllers/CalendarController.php
```diff
@@ -205,4 +209,16 @@ public function today(Calendar $calendar)
...
+    /**
+     * @param Campaign $campaign
+     * @param Calendar $calendar
+     * @return \Illuminate\Http\Resources\Json\AnonymousResourceCollection
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function eventLength(Calendar $calendar, ValidateLength $request)
+    {
+        $this->authorize('view', $calendar);
+        return $this->lengthValidatorService->validateLength($calendar, $request);
```
ilestis, line 222: You can use `return response()->json($this->lengthValidator...)` for a cleaner separation of concern

app/Http/Requests/ValidateLength.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace App\Http\Requests;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+class ValidateLength extends FormRequest
```
ilestis, line 7: Please use `ValidateReminderLength` to avoid class name collision if one day we have other Length validators

resources/js/crud.js
```diff
@@ -252,6 +252,28 @@ function registerEntityCalendarModal() {
         entityCalendarSubForm.fadeIn();
         loadCalendarDates(entityCalendarField.val());
     }
+
+    $('.entity-calendar-subform input[name="length"]').focusout(function () {
+        let url = $(this).data('url').replace('/0/', '/' + entityCalendarField.val() + '/')
```
ilestis, line 257: Can we avoid hitting the api when the field is empty to preserve server resources? :)

resources/js/crud.js
```diff
@@ -252,6 +252,28 @@ function registerEntityCalendarModal() {
...
+            day: entityCalendarDayField.val(),
+            month: entityCalendarMonthField.val(),
+            year: entityCalendarYearField.val(),
+            length: $('.entity-calendar-subform input[name="length"]').val(),
+        }
+      
+        $.ajax(url, {data: params}).done(function (data) {
+            data = JSON.parse(data);
+            if (data.overflow == true) {
+                console.log('hi');
```
ilestis, line 269: Cleanup logs

resources/js/crud.js
```diff
@@ -252,6 +252,28 @@ function registerEntityCalendarModal() {
...
+      
+        let params = {
+            day: entityCalendarDayField.val(),
+            month: entityCalendarMonthField.val(),
+            year: entityCalendarYearField.val(),
+            length: $('.entity-calendar-subform input[name="length"]').val(),
+        }
+      
+        $.ajax(url, {data: params}).done(function (data) {
+            data = JSON.parse(data);
```
ilestis, line 267: This won't be needed with switching to laravel's json responses

routes/campaign.php
```diff
@@ -131,6 +131,7 @@
 Route::get('/calendars/{calendar}/month-list', 'CalendarController@monthList')->name('calendars.month-list');
 Route::get('/calendars/{calendar}/events', 'CalendarController@events')->name('calendars.events');
 Route::get('/calendars/{calendar}/today', 'CalendarController@today')->name('calendars.today');
+Route::get('/calendars/{calendar}/validate-length', [\App\Http\Controllers\CalendarController::class, 'eventLength'])->name('calendars.event.length');
```
ilestis, line 134: Please use `-` instead of `.`, as we use `.` for subfolders that end up as `/` in the url

ilestis (changes requested):
Logic works, nice job! Code needs some cleanup to avoid headaches in the future :)

resources/js/crud.js
```diff
@@ -253,27 +253,28 @@ function registerEntityCalendarModal() {
...
-            data = JSON.parse(data);
-            if (data.overflow == true) {
-                console.log('hi');
-                $('.length-warning').show();
-            } else {
-                $('.length-warning').hide();
+    if ($('.entity-calendar-subform input[name="length"]').val()) {
+
+        $('.entity-calendar-subform input[name="length"]').focusout(function () {
+            let url = $(this).data('url').replace('/0/', '/' + entityCalendarField.val() + '/')
```
ilestis, line 259: Missing the don't do an ajax request if there is no content

resources/js/crud.js
```diff
@@ -253,27 +253,28 @@ function registerEntityCalendarModal() {
...
+        
+            let params = {
+                day: entityCalendarDayField.val(),
+                month: entityCalendarMonthField.val(),
+                year: entityCalendarYearField.val(),
+                length: $('.entity-calendar-subform input[name="length"]').val(),
```
ilestis, line 265: You can replace the selector with `$(this).val()`

ilestis (changes requested):
Missing one change, and some code cleanup

ilestis (approved):

PR #725, opened 2023-08-22
API: Entity Recovery

spitfire305:


app/Http/Controllers/Api/v1/EntityRecoveryApiController.php
```diff
@@ -0,0 +1,54 @@
...
+use App\Models\Campaign;
+use App\Models\Entity;
+use App\Http\Resources\EntityResource as Resource;
+use App\Http\Requests\RecoverEntity as Request;
+use App\Services\RecoveryService;
+
+class EntityRecoveryApiController extends ApiController
+{
+    /** @var RecoveryService */
+    protected $service;
```
ilestis, line 14: You can type the service and remove the comment :-)
`protected RecoveryService $service;`

resources/api-docs/1.0/entities.md
```diff
@@ -192,3 +195,86 @@ With each request to an object (ie. `character`, `location`, etc), you can inclu
...
+
+| Parameter | Type | Description
+| :- | :- | :- |
+| `entities` | `array` | The ids of the entities to recover. |
+
+### Examples
+
+| Method | Body | Headers |
+| :- |   :-   |  :-  |
+| POST | `"entities": [2,132,10]` | Default |
```
ilestis, line 272: We don't usually provide examples since we expect people to already know how to interact with an API. However the `Method|Uri|Headers` table is missing before the body. Please reproduce the doc layout from other files to keep it consistent.

resources/api-docs/1.0/entities.md
```diff
@@ -192,3 +195,86 @@ With each request to an object (ie. `character`, `location`, etc), you can inclu
...
+
+### Result
+
+```json
```
ilestis, line 276: Same with the success, we don't need an actual copy-paste, reuse the same layout in the rest of the API docs.

ilestis (changes requested):
Small tweaks to make the doc have the same formatting as the rest of kanka's api docs

PR #734, opened 2023-08-29
Boost Cooldown

spitfire305:


resources/views/settings/premium/remove.blade.php
```diff
@@ -11,18 +11,21 @@
     <h4 class="mt-0">
         {{ __('settings/premium.remove.title') }}
     </h4>
+    @if(!$boost->created_at->isBefore(Carbon\Carbon::now()->subDays(7)))
```
ilestis, line 14: Can you put this in a function on the Boost model, so that when we add boost manipulation on the API, we don't forget about this (hopefully) :D

ilestis (changes requested):

ilestis (approved):

PR #735, opened 2023-08-29
Conversation: don't allow adding same user multiple times

spitfire305:


ilestis (approved):
Was so confused just looking at the code, but it works!

PR #737, opened 2023-09-05
PayPal: Upgrade Sub

spitfire305:


app/Http/Controllers/Settings/SubscriptionController.php
```diff
@@ -91,6 +91,10 @@ public function change(Request $request)
         $isYearly = $period === 'yearly';
         $hasPromo = $isYearly && \Carbon\Carbon::create(2022, 10, 31)->isFuture();
         $limited = $this->subscription->isLimited();
+        // @phpstan-ignore-next-line
+        if ($user->subscribed('kanka') && str_contains($user->subscriptions()->first()->stripe_price, 'paypal')) {
```
ilestis, line 95: This same test is copy-pasted in many places. Please make it a single function on the User model

ilestis (changes requested):
Subscription page still doesn't load for a new account

ilestis (approved):

PR #740, opened 2023-09-12
Live Attributes: Data-attribute for current value

spitfire305:


resources/js/attributes.js
```diff
@@ -185,6 +185,11 @@ const listenToLiveForm = () => {
 
             let target = $('[data-uid="' + result.uid + '"]');
             //console.log('looking for', '[data-uid="' + result.uid + '"]', target);
+            if (result.checkbox == null) {
```
ilestis, line 188: We already do this if/else in PHP and now it's repeated in JS. Please refactor to only have the "logic" in one place, preferably PHP, by using a single variable ;-)

ilestis (changes requested):
Can be simplified

ilestis (approved):

PR #743, opened 2023-09-19
Pro rata in confirm page

spitfire305:


app/User.php
```diff
@@ -496,4 +496,77 @@ public function hasPayPal(): bool
...
+                $price = "10.00";
+            }
+        } elseif ($tier === 'Elemental') {
+            $price = "275.00";
+            if ($period == 'monthly') {
+                $price = "25.00";
+            }
+        }
+
+        if ($this->isSubscriber()) {
```
ilestis, line 532: That's a big if, could flip it around `if (!$this->isSubscriber() { return $price; }`

app/User.php
```diff
@@ -496,4 +496,77 @@ public function hasPayPal(): bool
...
+                }
+            }
+            if ($period == 'yearly') {
+                // @phpstan-ignore-next-line
+                $price = floatval($price) - ($oldPrice + ((floatval($price) / 365) * $this->subscriptions()->first()->updated_at->diffInDays(Carbon::now())));
+            } elseif ($monthly && $period == 'monthly') {
+                // @phpstan-ignore-next-line
+                $price = floatval($price) - ($oldPrice + ((floatval($price) / 31) * $this->subscriptions()->first()->updated_at->diffInDays(Carbon::now())));
+            } else {
+                return 'test';
```
ilestis, line 564: Will this ever show up? If it does, the code needs some comments to make it easier to parse like `// This happens when the user is born on a fullmoon`

resources/views/settings/subscription/change.blade.php
```diff
@@ -1,3 +1,6 @@
+@php
+    $upgrade = $user->upgradePrice($period, $tier);
```
ilestis, line 2: This should be prepared and sent in the controller. It also should be in the Subscription service rather than on the user, otherwise we're loading that function on every single page into memory when we only need it in one place :)

app/User.php
```diff
@@ -496,4 +496,77 @@ public function hasPayPal(): bool
         // @phpstan-ignore-next-line
         return $this->subscribed('kanka') && $this->subscriptions()->first() && str_contains($this->subscriptions()->first()->stripe_price, 'paypal');
     }
+
+    public function upgradePrice(string $period, string $tier): string
```
ilestis, line 500: Needs to be in either SubscriptionService, or t's own new service for an even cleaner code

ilestis (changes requested):
Small tweaks to keep the code clean

ilestis (approved):

PR #745, opened 2023-10-05
API: Entity Transform Endpoint

spitfire305:


app/Http/Controllers/Api/v1/EntityTransformApiController.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function transform(Request $request, Campaign $campaign)
+    {
+        $this->authorize('access', $campaign);
+
+        foreach ($request->entities as $id) {
```
ilestis, line 28: No permission check. I can transform as a user ALL entities of a campaign now, despite having 0 permissions to edit anything :-) Works on a public campaign I'm not a member of too

ilestis (changes requested):

ilestis (approved):

PR #746, opened 2023-10-06
API: Transfer to Campaign Endpoint

spitfire305:


app/Http/Controllers/Api/v1/EntityMoveApiController.php
```diff
@@ -0,0 +1,51 @@
...
+    public function transfer(Request $request, Campaign $campaign)
+    {
+        $this->authorize('access', $campaign);
+        $count = 0;
+        $copy = null === $request->copy ? true : $request->copy;
+
+        foreach ($request->entities as $id) {
+            $entity = Entity::find($id);
+            if ($this->authorize('update', $entity->child)) {
+                $this->service
```
ilestis, line 33: Ugly error when trying to copy to a campaign I'm not a part of. Missing a try/catch with a TranslatableException catch

resources/api-docs/1.0/entities.md
```diff
@@ -215,6 +216,25 @@ You can post an array with the ids of the entities you want to transform to the
...
+
+| Method | URI | Headers |
+| :- |   :-   |  :-  |
+| POST | `transfer` | Default |
+
+| Parameter | Type | Description
+| :- | :- | :- |
+| `entities` | `array`(required) | The ids of the entities to transfer or copy. |
+| `campaign_id` | `integer`(required) | The id of the campaign the entity will be transfered or copied to. |
+| `copy` | `boolean` | True if the entity will be copied, false if the entity will be transfered, defaults to copy if left empty |
```
ilestis, line 232: This is really confusing. The endpoint is called "transfer", but if you don't provide a copy param, it will copy? I know I said to copy the original way ti works, but this makes no sense. Copy should be optional and only copy if set to true. Also missing the end `.` to keep it consistent.

ilestis (changes requested):

app/Http/Controllers/Api/v1/EntityMoveApiController.php
```diff
@@ -39,13 +40,21 @@ public function transfer(Request $request, Campaign $campaign)
...
-        }
 
-        if ($copy) {
-            return response()->json(['success' => 'Succesfully copied ' . $count . ' entities.']);
+            if ($copy) {
+                return response()->json(['success' => 'Succesfully copied ' . $count . ' entities.']);
+            }
+            return response()->json(['success' => 'Succesfully transfered ' . $count . ' entities.']);
+        
+        } catch (Exception $e) {
```
ilestis, line 52: Needs a catch for TranslatableException, and for standard exception to be handled by the rest of the app. If it's a standard exception, I need to get an email that something went wrong with the code :)

app/Http/Controllers/Api/v1/EntityMoveApiController.php
```diff
@@ -25,12 +26,12 @@ public function transfer(Request $request, Campaign $campaign)
     {
         $this->authorize('access', $campaign);
         $count = 0;
-        $copy = null === $request->copy ? true : $request->copy;
-
-        foreach ($request->entities as $id) {
-            $entity = Entity::find($id);
-            if ($this->authorize('update', $entity->child)) {
-                $this->service
+        $copy = is_null($request->copy) ? false : $request->copy;
```
ilestis: Can be simplified to `$copy = $request->has('copy');`

ilestis (changes requested):

ilestis (approved):

PR #748, opened 2023-10-09
API: Bug updating entity removes image uuid

spitfire305:


app/Observers/EntityObserver.php
```diff
@@ -227,7 +227,7 @@ public function saveBoosted(Entity $entity): void
         // Gallery is now available to all
         if (request()->has('entity_image_uuid')) {
             $entity->image_uuid = request()->get('entity_image_uuid');
-        } else {
+        } elseif (!request()->is('1.0/*')) {
```
ilestis, line 230: This seems very hard-coded (what if we introduce a 2.0 api next year?). Should be `Domain::isApp()` instead to make sure we are not in the API subdomain

ilestis (changes requested):

ilestis (approved):

PR #751, opened 2023-10-11
Type is now clickable

spitfire305:


app/Models/Entity.php
```diff
@@ -438,4 +438,19 @@ public function postPositionOptions($position = null): array
...
+
+    /**
+     * Type filter link
+     */
+    public function typeLink(string $displayName = null): string
+    {
+        $campaign = CampaignLocalization::getCampaign();
+        $defaultOptions = auth()->check() && auth()->user()->entityExplore === '1' ? [$campaign, 'm' => 'table'] : [$campaign];
+
+        return '<a href="' .
```
ilestis, line 450: I try to avoid writing html in models. You can use `return link_to_route()` instead, and let laravel figure out how to write the html :)

ilestis (changes requested):

ilestis (approved):

PR #754, opened 2023-10-17
Refactored parent_location_id to parent_location and fixed sidebar issue

spitfire305:


resources/views/cruds/fields/location.blade.php
```diff
@@ -12,7 +12,7 @@
 @endphp
 <x-forms.foreign
     :campaign="$campaign"
-    :name="isset($isParent) ? 'parent_location_id' : 'location_id'"
+    :name='"location_id"'
```
ilestis, line 15: Confused by the quotes? If passing a string, just `name="location_id"` is enough, `:` is for a php property.

ilestis (changes requested):
Tiniest change.

ilestis (approved):

PR #755, opened 2023-10-18
Export: Include Gallery

spitfire305:


app/Traits/ExportableTrait.php
```diff
@@ -31,10 +31,14 @@ public function export(): string
         // Entity values
         if (!empty($this->entity)) {
             // Todo: put these in with()
-            $foreigns = ['posts', 'relationships', 'abilities', 'events', 'tags', 'assets', 'entityAttributes'];
+            $foreigns = ['posts', 'relationships', 'abilities', 'events', 'tags', 'assets', 'entityAttributes', 'image', 'header'];
```
ilestis, line 34: Missing eager loading to not crash mysql on a large campaign :-) You can temporarily set `APP_LAZY=true` in your .env to see what needs to be added.

ilestis (changes requested):
Missing performance tweaks to make it work on campaigns larger than 50 entities

app/Services/Campaign/ExportService.php
```diff
@@ -142,27 +144,19 @@ protected function gallery(): self
...
     {
-        $route = $path . '/' . $image->name;
-
-        if ($image->imageFolder) {
-            $route = $route . '/';
-            $route = $this->processFolder($image->imageFolder, $route);
-        } else {
-            $route = $route . '.' . $image->ext;
+        if(!$this->assets) {
+            $this->archive->addFromString('gallery/' . Str::slug($image->name) . '.json', $image->export());
```
ilestis, line 150: I have multiple images called "screenshot" which is resulting in lost data. Please include the model's uuid or id (not sure if we have ids on this?) in the file name :to prevent data loss :-)

app/Services/Campaign/ExportService.php
```diff
@@ -142,27 +144,19 @@ protected function gallery(): self
...
+    protected function processImage($image): self
     {
-        $route = $path . '/' . $image->name;
-
-        if ($image->imageFolder) {
-            $route = $route . '/';
-            $route = $this->processFolder($image->imageFolder, $route);
-        } else {
-            $route = $route . '.' . $image->ext;
+        if(!$this->assets) {
```
ilestis, line 149: Return early to avoid code being nested

app/Services/Campaign/ExportService.php
```diff
@@ -142,27 +144,19 @@ protected function gallery(): self
...
-        return $route;
-    }
-
-    protected function processImage($image): self
-    {
         // Boosted image?
         if (!$image->is_folder) {
-            $path = $this->processFolder($image, 'gallery/');
-            $this->archive->addFromString($path, Storage::get($image->path));
+            $this->archive->addFromString('gallery/' . $image->name . '.' . $image->ext, Storage::get($image->path));
```
ilestis, line 157: Issue with my 30 images called "screenshot.png"

ilestis (changes requested):

app/Services/Campaign/ExportService.php
spitfire305: Its set the same way as the already existing process function, I dont see a way of returning earlier.

ilestis (changes requested):
Forgot to restart the queue

PR #757, opened 2023-10-27
Export logs

spitfire305:


app/Services/Campaign/ExportService.php
```diff
@@ -46,13 +48,28 @@ public function queue(): self
         $this->campaign->export_date = date('Y-m-d');
         $this->campaign->saveQuietly();
 
-        Export::dispatch($this->campaign, $this->user, false);
-        Export::dispatch($this->campaign, $this->user, true);
+        $entitiesExport = CampaignExport::create([
```
ilestis, line 51: I feel like these should be created in the Job rather than here and passed as a parameter

ilestis (changes requested):
Where is the job updated based on the status? I don't see the size being saved in the db either. We also need the path of the export in the db for the next ticket :-)

app/Services/Campaign/ExportService.php
```diff
@@ -46,13 +48,28 @@ public function queue(): self
...
+            'status' => CampaignExport::STATUS_SCHEDULED,
+        ]);
+
+        Export::dispatch($this->campaign, $this->user, $assetExport, true);
 
         return $this;
     }
 
-    public function export(): self
+    public function export(): int
```
ilestis, line 72: export() returning an int smell of base code. Make a new function size() to get the size of the export.

database/migrations/2023_10_25_173253_create_campaign_exports_table.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        Schema::create('campaign_exports', function (Blueprint $table) {
+
+            $table->increments('id');
+            $table->unsignedInteger('campaign_id');
+            $table->unsignedInteger('created_by')->nullable();
+            $table->integer('size');
+            $table->tinyInteger('type');
+            $table->tinyInteger('status');
+            $table->timestamps();
```
ilestis, line 21: Missing path string for the next ticket

ilestis (changes requested):

ilestis (approved):

PR #758, opened 2023-10-31
Campaign Export: Datagrid

spitfire305:


app/Renderers/Layouts/Campaign/CampaignExport.php
```diff
@@ -0,0 +1,109 @@
...
+     * @return array[]
+     */
+    public function columns(): array
+    {
+        $columns = [
+            'created_by' => [
+                'key' => 'user.name',
+                'label' => 'campaigns.exports.created_by',
+                'render' => function ($model) {
+                    $html = '<a class="block break-all truncate" href="' . route('users.profile', [$model->user]) . '" target="_blank">' . $model->user->name . '</a>';
```
ilestis, line 21: Crashes if the user who created the export deletes their account, or if the export was triggered by a command line `sail artisan campaigns:export <id>` ;-)

app/Renderers/Layouts/Campaign/CampaignExport.php
```diff
@@ -0,0 +1,109 @@
...
+                    return $html;
+                },
+            ],
+            'size' => [
+                'label' => 'campaigns.exports.size',
+                'render' => function ($model) {
+                    if ($model->size == 0) {
+                        return '<1 MB';
+                    }
+                    return $model->size . ' MB';
```
ilestis, line 31: Hopefully it never happens, but adding `number_format()` would help here in case there is ever a >1000MB export.

app/Renderers/Layouts/Campaign/CampaignExport.php
```diff
@@ -0,0 +1,109 @@
...
+                },
+            ],
+            'download' => [
+                'label' => 'campaigns.exports.download',
+                'render' => function ($model) {
+                    if ($model->path && Storage::exists($model->path)) {
+                        $html = '<a class="block break-all truncate" href="' . Storage::url($model->path) . '" target="_blank">' . __('campaigns.exports.download') . '</a>';
+                        return $html;
+                    } elseif ($model->path) {
+                        return __('campaigns.exports.expired');
```
ilestis, line 84: Adding a `<span class="text-neutral-content">` around the expired text makes it visually less eye-catching

app/Models/CampaignExport.php
```diff
@@ -38,6 +41,14 @@ class CampaignExport extends Model
         'path',
     ];
 
+    public $sortable = [
+        'status',
+        'type',
+        'updated_at',
```
ilestis, line 47: Updated_at isn't in the datagrid and thus doesn't need to be sortable

app/Renderers/Layouts/Campaign/CampaignExport.php
```diff
@@ -0,0 +1,109 @@
...
+                },
+            ],
+
+            'created_at' => [
+                'key' => 'created_at',
+                'label' => 'campaigns.exports.created',
+                'render' => function ($model) {
+                    $html = '';
+
+                    if (!empty($model->created_at)) {
```
ilestis, line 71: Confused how a model would never have a created_at value. But then the code is faster to read 
```php
if (empty($model->created_at)) {
   return '';
   }
   // the code
   ```

File: lang/en/campaigns.php [hunk 6]
```diff
@@ -468,4 +468,19 @@
         'public'    => 'Public campaign',
         'review'    => 'Awaiting Review',
     ],
+    'exports'                      => [
+        'created_by'      => 'Created by',
```
ilestis, line 472: Please use `campaigns.members.fields.name` instead

lang/en/campaigns.php
```diff
@@ -468,4 +468,19 @@
...
+        'size'            => 'Size',
+        'type'            => 'Type',
+        'status'          => 'Status',
+        'status_finished' => 'Finished',
+        'status_scheduled'=> 'Scheduled',
+        'status_failed'   => 'Failed',
+        'status_running'  => 'Running',
+        'type_assets'     => 'Assets',
+        'type_entities'   => 'Entities',
+        'created'         => 'Created',
```
ilestis, line 482: Created is already translated in `campaigns.invites.fields.created`, please don't add more work for our poor translators :)

resources/views/campaigns/export.blade.php
```diff
@@ -36,5 +36,16 @@
             {{ __('campaigns/export.errors.limit') }}
         </x-alert>
         @endif
+
+        <?php /** @var \App\Models\Campaign $campaign
```
ilestis, line 40: Don't think this block is needed

ilestis (changes requested):
Small changes

ilestis (approved):

PR #763, opened 2023-11-02
Fixed Bug

spitfire305:


resources/views/calendars/events/_subform.blade.php
```diff
@@ -22,6 +22,7 @@
     <x-forms.field field="comment" css="col-span-2" :label="__('calendars.fields.comment')">
         {!! Form::text('comment', null, ['placeholder' => __('calendars.placeholders.comment'), 'maxlength' => 191]) !!}
     </x-forms.field>
+    <span id="entity-calendar-modal-add">
```
ilestis, line 25: In HTML, a span is an inline element. If you want a block, `div` is required for semantically correct HTML

ilestis (changes requested):

ilestis (approved):

PR #764, opened 2023-11-07
Added parent and children to API resource

spitfire305:


app/Http/Resources/EntityResource.php
```diff
@@ -203,9 +203,23 @@ public function entity(array $prepared = [])
             $merged['entity_abilities'] = EntityAbilityResource::collection($misc->entity->abilities);
             $merged['entity_assets'] = EntityAssetResource::collection($misc->entity->assets);
         }
-
+        if ($misc->ancestors) {
```
ilestis, line 206: Let's only load these if related is set to true, to avoid it loading by default. Also can you make it work with APP_LAZY=true?

ilestis (changes requested):

app/Http/Resources/EntityResource.php
spitfire305: Both $misc->ancestors and $misc->getAncestors() work with APP_LAZY set to true

ilestis: image
I get an error about lazy on another relation directly when calling `locations?related=1` so I'm a bit confused how it's working on your end :)

ilestis: When I fix the missing with() for abilities, I now get the error for ancestors 
image

ilestis (approved):

PR #765, opened 2023-11-09
Meilisearch VIP

ilestis:


app/Services/Search/EntitySearchService.php
```diff
@@ -0,0 +1,86 @@
...
+    use CampaignAware;
+
+
+    /**
+     * Send search request
+     * @param string|null $query Search term
+     */
+    public function search(string $term = null): array
+    {
+        $client = new Client('http://meilisearch:7700', 'password');
```
ilestis, line 23: Why isn't this using scout? And everything is hardcoded?

app/Services/Search/EntitySearchService.php
```diff
@@ -0,0 +1,86 @@
...
+        $client->getKeys();
+        $ids = [];
+        $results = $client->index('entities')->search($term, ['filter' => 'campaign_id = ' . $this->campaign->id, 'attributesToRetrieve' => ['id', 'entity_id', 'type'], 'limit' => 20])->getHits();
+        $attributeIds = [];
+        $timelineElementIds = [];
+        $postIds = [];
+        $questElementIds = [];
+
+        foreach ($results as $result) {
```
ilestis, line 32: Split up the service into smaller functions so that I know what's going on just from the function name

app/Models/QuestElement.php
```diff
@@ -122,4 +124,49 @@ public function editingUsers()
...
+    {
+        return 'entities';
+    }
+
+    protected function makeAllSearchableUsing($query)
+    {
+
+        return $query
+        ->leftJoin('quests', 'quests.id', '=', 'quest_elements.quest_id')
+        ->select([$this->getTable() . '.*', 'entities.id as entity_id'])
```
ilestis: Weirded out by the select being in between two left joins

app/Models/Attribute.php
```diff
@@ -365,4 +367,43 @@ public function listRangeText(): string
...
+
+    protected function makeAllSearchableUsing($query)
+    {
+        return $query
+        ->leftJoin('entities', 'attributes.entity_id', '=', 'entities.id');
+    }
+
+    public function toSearchableArray()
+    {
+        $array = $this->toArray();
```
ilestis, line 396: Why do we need to do toArray() and not directly `return [ 'campaign_id' => $this->entity->campaign_id, etc]`? This seems like a waste of processing power to make everything an array (also makes ctrl+click to go to the definition not work in phpstorm)

config/scout.php
```diff
@@ -133,9 +134,9 @@
         'host' => env('MEILISEARCH_HOST', 'http://localhost:7700'),
         'key' => env('MEILISEARCH_KEY'),
         'index-settings' => [
-            // 'users' => [
-            //     'filterableAttributes'=> ['id', 'name', 'email'],
-            // ],
+            'entities' => [
+                'filterableAttributes'=> ['id', 'campaign_id'],
+            ],
```
ilestis, line 139: Missing a "docs/meilisearch.md" file to explain how to set things up. How do I batch add characters, locations, posts?

ilestis (commented):
I opened the PR like a dumb dumb so can't request changes

PR #770, opened 2023-11-21
Calendar: Show Birthdays

spitfire305:


ilestis (commented):
Death isn't being registered properly

app/Renderers/CalendarRenderer.php
```diff
@@ -902,10 +899,7 @@ protected function parseReminders(Collection $reminders): void
                     $this->recurring[$event->recurring_periodicity][] = $event;
                 }
             } else {
-                if ($event->type_id == EntityEventType::DEATH) {
-                    $this->deaths[$event->entity_id] = $event;
-                } elseif ($event->type_id == EntityEventType::BIRTH) {
-                    $this->births[$event->entity_id] = $event;
+                if ($event->isBirth() && $event->death && $event->death->isPastDate($this->getYear(), $event->month, $event->day)) {
```
ilestis, line 902: This check should happen way earlier in the loop ( https://dev.to/jpswade/return-early-12o5 ) and PHPStan needs to be fixed :)

ilestis (changes requested):

ilestis (approved):

PR #772, opened 2023-11-24
Email Cleanup

spitfire305:


app/Jobs/Emails/SubscriptionCreatedEmailJob.php
```diff
@@ -51,11 +51,12 @@ public function handle()
...
-        // Send an email to the admins
-        Mail::to('hello@kanka.io')
-            ->send(
-                new NewSubscriptionMail($user, $this->period, $this->new)
-            );
+        if ($this->new) {
+            // Send an email to the admins
+            Mail::to('hello@kanka.io')
+                ->send(
+                    new NewSubscriptionMail($user, $this->period, $this->new)
```
ilestis, line 58: You removed the new param but are still passing it to the job

ilestis (changes requested):
It's still sending the "failed too many times" email. I think it's called Cancelled instead of failed, but the email subject is explicit "failed too many times" which I don't want to receive anymore

ilestis (approved):

PR #776, opened 2023-12-01
Add multiple Abilities

spitfire305:


app/Models/Ability.php
```diff
@@ -205,27 +205,29 @@ public function entryWithAttributes()
...
-        $entity = Entity::with('abilities')->findOrFail($entityId);
-
-        // Make sure the tag isn't already attached to the entity
-        foreach ($entity->abilities as $ability) {
-            if ($ability->ability_id == $this->id) {
-                return true;
+        $entityIds = Arr::get($request, 'entities');
+        $entities = Entity::with('abilities')->findOrFail($entityIds);
+        $count = 0;
+        foreach ($entities as $entity) {
```
ilestis, line 215: Could we not replace all of this function with 
```php
$visibility = Arr::get($request, 'visibility_id', \App\Enums\Visibility::All);
$sync = [];
foreach ($entities as $entity) {
   $sync[$entity->id] = ['visibility_id' => $visibility];
}
$this->entities->sync($sync)
```
?

or does that lose information already saved on attached abilities like the "notes" and "position"?

resources/views/abilities/entities/_form.blade.php
```diff
@@ -2,13 +2,9 @@
...
 
 <x-grid type="1/1">
-    @include('cruds.fields.entity', [
-        'required' => true,
-        'route' => 'search.ability-entities',
-        'placeholder' => __('entities/relations.placeholders.target'),
-        'preset' => false,
-        'dropdownParent' => request()->ajax() ? '#primary-dialog' : null,
-    ])
+    <x-forms.field field="entities" :required="true" >
```
ilestis, line 5: Highly recommend making `route` a param otherwise you'll ahve a really bad time with the other tickets

resources/views/abilities/entities/_form.blade.php
```diff
@@ -2,13 +2,9 @@
...
+        @include('components.form.entities', ['options' => ['exclude-entity' => $model->entity->id]])
```
ilestis, line 6: components shouldn't be included in blade views but use laravel's components feature <x-entities>

ilestis (changes requested):

resources/views/abilities/entities/_form.blade.php
```diff
@@ -2,14 +2,18 @@
...
-        </x-form.entities>
-    </x-forms.field>
+    <x-forms.foreign
+        field="entities"
+        :required="true"
+        label="abilities.show.tabs.entities"
+        multiple="multiple"
+        name="entities[]"
+        id="entities[]"
+        :options="['exclude-entity' => $model->entity->id, 'route' => 'search.ability-entities']"
```
ilestis, line 12: Please remove "options", every parameter is already an option.

ilestis (changes requested):

ilestis (approved):

PR #777, opened 2023-12-01
Tag children add multiple entities

spitfire305:


app/Http/Controllers/Tags/ChildController.php
```diff
@@ -79,8 +79,8 @@ public function store(StoreTagEntity $request, Campaign $campaign, Tag $tag)
             $redirectUrlOptions['tag_id'] = $tag->id;
         }
 
-        $tag->attachEntity($request->only('entity_id'));
+        $count = $tag->attachEntity($request->only('entities'));
```
ilestis, line 82: attachEntitie__s__() please :)

app/Models/Tag.php
```diff
@@ -292,26 +292,28 @@ public function hasColour(): bool
...
             }
-        }
 
-        $entityTag = EntityTag::create([
-            'tag_id' => $this->id,
-            'entity_id' => $entityId
-        ]);
-
-        return $entityTag !== false;
+            EntityTag::create([
```
ilestis, line 310: Can we just do `$entity->tags->syncWithoutDetatching([$this->id])` ?

spitfire305: syncWithoutDetatching() doesnt work with hasManyThrough relations.

resources/views/components/forms/foreign.blade.php
```diff
@@ -19,9 +22,14 @@ class="w-full select2 join-item"
...
             data-allow-clear="{{ $allowClear ? 'true' : 'false' }}"
             @if (!empty($dropdownParent)) data-dropdown-parent="{{ $dropdownParent }}" @endif
+            :dropdownParent="request()->ajax() ? '#primary-dialog' : null"
     >
         @foreach ($options as $key => $value)
-            <option value="{{ $key }}">{!! $value !!}</option>
+            @if ($multiple)
+                <option value="{{ $key }}">{!! $value !!}</option>
+            @else
+                <option value="{{ $key }}" class="select2-entity" selected="selected">{{ $entity->name }}</option>
```
ilestis, line 31: Why $entity->name ? Where is that value coming from magically? Why the select2-entity class? When editing a location, does this change break the parent location dropdown?

resources/views/tags/entities/_form.blade.php
```diff
@@ -1,14 +1,17 @@
...
-            :campaign="$campaign"
-            :ajax="request()->ajax()"
-        >
-        </x-form.entities>
-    </x-forms.field>
+    <x-forms.foreign
+        field="entities"
+        :required="true"
+        label="abilities.show.tabs.entities"
+        multiple="multiple"
```
ilestis, line 8: I feel like multiple should be a boolean, and validated on the Foreign.php component

resources/views/components/forms/foreign.blade.php
```diff
@@ -10,7 +10,10 @@
         <div class="join w-full">
     @endif
 
-    <select name="{{ $name }}" id="{{ $id }}"
+    <select 
+            multiple="$multiple"
```
ilestis, line 14: image
I can now set multiple locations on a family

ilestis (changes requested):
Broke all other dropdowns

ilestis (approved):
:chefkiss:

PR #780, opened 2023-12-05
Bulk Relation Creation

spitfire305:


app/Http/Controllers/Api/v1/EntityRelationApiController.php
```diff
@@ -40,26 +48,14 @@ public function store(Request $request, Campaign $campaign, Entity $entity)
         $this->authorize('access', $campaign);
         $this->authorize('update', $entity->child);
 
+        [$new, $count] = $this->relationService->campaign($campaign)->createRelations($request);
+
         if ($request->has('targets')) {
```
ilestis, line 53: No longer needed? It's in the service too

app/Services/Entity/RelationService.php
```diff
@@ -0,0 +1,44 @@
...
+            $count++;
+            if (!isset($new)) {
+                $new = $relation;
+            }
+            if ($request->has('two_way')) {
+                $relation->createMirror();
+                $count++;
+            }
+        }
+        $results = [$new, $count];
```
ilestis, line 41: I really really dislike returning array with mixed data, it makes the code really clumsy. I would prefer that both are set in variables on the service, and with a getter function called in the controller when needed, thanks.

app/Http/Controllers/Api/v1/EntityRelationApiController.php
```diff
@@ -40,26 +48,14 @@ public function store(Request $request, Campaign $campaign, Entity $entity)
...
-
-        foreach ($entities as $entity_id) {
-            $data['target_id'] = $entity_id;
-            $model = Relation::create($data);
-            if ($request->has('two_way')) {
-                $model->createMirror();
-            }
-        }
-
         return Resource::collection($entity->relationships()->has('target')->whereIn('target_id', $entities)->paginate());
```
ilestis, line 59: PHPStan is unhappy

ilestis (changes requested):

PR #785, opened 2023-12-12
Added a Connection filter and fixed a bug

spitfire305:


app/Datagrids/Filters/CharacterFilter.php
```diff
@@ -75,6 +75,7 @@ public function build()
             ->hasAttributes()
             ->tags()
             ->attributes()
+            ->connection()
```
ilestis, line 78: All the other filters are plural, so would be nice to keep the same nomenclature :)
Also needs to be added to Locations, Families, Tags, Calendars, etc

app/Models/Character.php
```diff
@@ -462,6 +462,9 @@ public function filterableColumns(): array
             'race_id',
             'family_id',
             'races',
+            'connection',
```
ilestis, line 465: Not needed here since it's in the HasFilter trait's `defaultFilterableColumns`

lang/en/entities/relations.php
```diff
@@ -54,6 +54,10 @@
         'two_way'           => 'Mirror connection',
         'unmirror'          => 'Unmirror this connection.',
     ],
+    'filters'           => [
+        'name'          => 'Name',
```
ilestis, line 58: The screenshot on the ticket shows other labels for the fields :-)

ilestis (changes requested):
Wrong labels and missing on Locations, Tags, Families, Orgs etc

ilestis (approved):

PR #788, opened 2023-12-13
Update to PHP 8.3

spitfire305:


docker-compose.yml
```diff
@@ -3,11 +3,11 @@ version: '3'
 services:
     laravel.test:
         build:
-            context: ./docker/8.2
+            context: ./vendor/laravel/sail/runtimes/8.3
```
ilestis, line 6: Our custom PHP.ini is now no longer loaded

ilestis (changes requested):

docker/8.0/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM ubuntu:20.04
```
ilestis, line 1: We don't need any of the 8.0 and 8.1 stuff, since we're only now using 8.3 :-)

ilestis (changes requested):

ilestis (approved):

PR #795, opened 2024-01-04
List Widget now links with filters

spitfire305:


app/Models/CampaignDashboardWidget.php
```diff
@@ -457,4 +457,25 @@ public function visible(): bool
         // Linked but no entity or no child? Permission issue or deleted entity
         return !empty($this->entity);
     }
+
+    public function filteredLink(string $entityString = null): string
+    {
+        $link = route('characters.index', [$this->campaign_id]);
```
ilestis, line 463: Feel like it will crash if the entity type is set to all :-)
image

ilestis (changes requested):

ilestis (approved):

PR #796, opened 2024-01-05
Entity List Widget: Filter on user.

spitfire305:


lang/en/dashboard.php
```diff
@@ -124,12 +124,14 @@
         ],
         'fields'                    => [
             'class'             => 'CSS class',
+            'created_by'        => 'Created by',
```
ilestis, line 127: Don't we already have this translated?

lang/en/dashboard.php
```diff
@@ -124,12 +124,14 @@
...
             'dashboard'         => 'Dashboard',
             'name'              => 'Custom widget name',
             'optional-entity'   => 'Link to entity',
             'order'             => 'Ordering',
             'size'              => 'Size',
             'text'              => 'Text',
+            'updated_by'        => 'Last updated by',
```
ilestis, line 134: Same here, but we don't need the field (see later)

resources/views/dashboard/widgets/forms/_recent.blade.php
```diff
@@ -80,6 +80,14 @@
             ], null, ['class' => '']) !!}
                 </x-forms.field>
                 @includeWhen(!empty($dashboards), 'dashboard.widgets.forms._dashboard')
+
+                <x-forms.field field="config[created_by]" :label="__('dashboard.widgets.fields.created_by')">
```
ilestis, line 84: We already have a filters field, we need to use that instead, by just adding the parameters there

ilestis (changes requested):
Please use the existing filters field instead of adding new fields

ilestis (approved):

PR #797, opened 2024-01-08
Assets: Allow CSV

spitfire305:


app/Rules/EntityFileRule.php
```diff
@@ -48,7 +48,7 @@ public function passes($attribute, $value)
             return false;
         }
 
-        return !(!in_array($value->getClientOriginalExtension(), ['mp3', 'ogg', 'json']))
+        return !(!in_array($value->getClientOriginalExtension(), ['mp3', 'ogg', 'json', 'csv']))
```
ilestis, line 51: Does csv detection not work on the guessExtension() line? Because getClientOriginalExtension is less secure (basically looks at the file name)

ilestis (changes requested):

PR #807, opened 2024-01-24
Email Validation.

spitfire305:


app/Jobs/Emails/Subscriptions/EmailValidationJob.php
```diff
@@ -0,0 +1,52 @@
...
+
+class EmailValidationJob implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    /** @var int user id */
+    protected $user;
```
ilestis, line 22: user should be cast as an int if it's an int :)

app/Jobs/Emails/Subscriptions/EmailValidationJob.php
```diff
@@ -0,0 +1,52 @@
...
+    protected $token;
```
ilestis, line 23: and token a string if it's a string

app/Mail/Subscription/User/ValidationEmail.php
```diff
@@ -0,0 +1,48 @@
...
+
+class ValidationEmail extends Mailable
+{
+    use Queueable;
+    use SerializesModels;
+
+    /**
+     * @var User
+     */
+    public $user;
```
ilestis, line 18: Same here, this can be cast instead of using the ugly /** @var **/ trick

app/User.php
```diff
@@ -433,6 +436,12 @@ public function isFrauding(): bool
         if (!empty($this->provider)) {
             return false;
         }
+
+        $validation = UserValidation::where('user_id', $this->id)->where('is_valid', true)->first();
```
ilestis, line 440: Since we are already on the user, it would be simpler to make it a relation on UserRelations , and is_valid into a scope `$this->userValidation->valid()->first()`

app/User.php
```diff
@@ -453,6 +462,35 @@ public function isFrauding(): bool
             ->count() >= 2;
     }
 
+    public function requiresEmail(): self
```
ilestis, line 465: This 30 line function is now loaded in memory on every single page load of Kanka, but is used in a single place. Please refactor to put it in a service.

database/migrations/2024_01_22_203110_create_user_validations_table.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('user_validations', function (Blueprint $table) {
+            $table->id();
+            $table->uuid('token');
+            $table->unsignedInteger('user_id');
+            $table->boolean('is_valid');
```
ilestis, line 17: Needs a default if not null

resources/views/emails/subscriptions/validation/user-html.blade.php
```diff
@@ -0,0 +1,22 @@
+@extends('emails.base', [
+    'utmSource' => 'subscription',
```
ilestis, line 2: Wrong tracking codes :-) jon will be unhappy

resources/views/emails/subscriptions/validation/user-html.blade.php
```diff
@@ -0,0 +1,22 @@
...
+@section('content')
+    <p>
+        <strong>Email Validation</strong>
+    </p>
+    <p>
+        {{ __('emails/subscriptions/upcoming.dear', ['name' => $user->name]) }},
+    </p>
+
+    <p>This is an automatic notification.</p>
+    <p>To validate the email for your Kanka account click <a href="{{ 'https://app.kanka.io/users/' . $user->id . '/validation?token=' . $token }}">here</a>. This link will expire in 24 hours.</p>
```
ilestis, line 15: I feel like the url should be a param set in the email rather than copy-pasted

resources/views/settings/subscription/change.blade.php
```diff
@@ -76,7 +79,7 @@
                 </li>
                 <li role="presentation">
                     <a href="#giropay" aria-controls="settings" role="tab" data-toggle="tab">
-                        giropay
+                        Giropay
```
ilestis, line 82: giropay was written correctly (no capitalisation)

app/Http/Controllers/User/EmailValidationController.php
```diff
@@ -0,0 +1,30 @@
...
+namespace App\Http\Controllers\User;
+
+use App\Http\Controllers\Controller;
+use App\Models\UserValidation;
+use App\User;
+use Illuminate\Http\Request;
+
+class EmailValidationController extends Controller
+{
+    public function validateEmail(Request $request, User $user)
```
ilestis, line 12: Email providers **will** open all links in emails to check for spam/fishing, so a user receiving the code will have very weird stuff happen.

The page should only work if the user is logged in to avoid this.

ilestis (changes requested):

app/Http/Controllers/User/EmailValidationController.php
```diff
@@ -11,6 +11,10 @@ class EmailValidationController extends Controller
 {
     public function validateEmail(Request $request, User $user)
     {
+        if (!(auth()->check() && auth()->user()->id == $user->id)) {
```
ilestis, line 14: The first check can be done with a middleware on the controller's __construct instead

ilestis (changes requested):

ilestis (approved):

app/User.php
```diff
@@ -433,6 +433,12 @@ public function isFrauding(): bool
         if (!empty($this->provider)) {
             return false;
         }
+
+        $validation = $this->userValidation->valid()->first();
```
ilestis, line 437: This crashes and makes no sense

resources/views/settings/subscription/change.blade.php
```diff
@@ -5,9 +5,14 @@
 <article class="text-center max-w-xl container">
 
     <x-grid type="1/1">
-    @if ($user->isFrauding())
+    @if (!$user->isFrauding())
+        @inject('emailService', 'App\Services\Users\EmailValidationService')
```
ilestis, line 9: This should be in the controller

ilestis (changes requested):
Multiple issues that need fixing

app/Http/Controllers/Settings/SubscriptionController.php
```diff
@@ -93,6 +97,8 @@ public function change(Request $request, Tier $tier)
         }
         $upgrade = $this->subscriptionUpgrade->user($user)->tier($tier)->upgradePrice($period);
         $currency = $user->currencySymbol();
+        
+        $this->emailValidation->user($user)->requiresEmail();
```
ilestis, line 101: This is now sending the email to every single user opening the subscription page, while it should only send to people suspected of frauding _when_ they try and subscribe

app/User.php
```diff
@@ -426,15 +426,15 @@ public function hasUnread(): bool
     public function isFrauding(): bool
     {
         // Fraud detection can be turned on or off
-        if (!config('subscription.fraud_detection')) {
+        if (config('subscription.fraud_detection')) {
```
ilestis, line 429: Eeeehhhh :-) please don't change the logic

ilestis (changes requested):
Still not working as expected

resources/views/settings/subscription/change.blade.php
```diff
@@ -5,7 +5,10 @@
 <article class="text-center max-w-xl container">
 
     <x-grid type="1/1">
-    @if (!$user->isFrauding())
+    @if ($validationService)
```
ilestis, line 8: The code was correct before this change

ilestis (changes requested):
Fixed one thing, broke another

ilestis (approved):

PR #809, opened 2024-01-26
Premium: Public but not visible in public.

spitfire305:


app/Services/Api/CampaignService.php
```diff
@@ -105,13 +105,14 @@ protected function campaigns(): self
     {
         $this->data['campaigns'] = [];
 
-        if ($this->isDefaultRequest()) {
+        if (!$this->isDefaultRequest()) {
```
ilestis, line 108: This smells fishy

app/Services/Campaign/BoostService.php
```diff
@@ -121,8 +121,11 @@ public function unboost(CampaignBoost $campaignBoost): self
             }
             $this->user->log(UserLog::TYPE_CAMPAIGN_UNBOOST);
         }
-
-        $this->campaign->boost_count = $this->campaign->boosts()->count();
+        $boostCount = $this->campaign->boosts()->count();
+        $this->campaign->boost_count = $boostCount;
+        if ($boostCount == 0 && $this->campaign->isDiscreet()) {
```
ilestis, line 126: No need for the if, since it's not possible to downgrade from super to boosted

ilestis (changes requested):

ilestis (approved):

PR #810, opened 2024-01-31
Discord notification about new features

spitfire305:


app/Jobs/Discord/SendNewFeature.php
```diff
@@ -0,0 +1,79 @@
...
+
+class SendNewFeature implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    /**  */
+    protected int $feature;
```
ilestis, line 22: Feature isn't an int... but it should be (see other PR)

config/discord.php
```diff
@@ -16,6 +16,9 @@
     'bot_secret' => getenv('DISCORD_BOT_SECRET'),
     'bot_token' => getenv('DISCORD_BOT_TOKEN'),
 
+    'webhook' => env('DISCORD_WEBHOOK', 'null'),
```
ilestis, line 19: Make webhook an array with a proper key so that in the future when we add more webhooks it doesn't become a mess :)

ilestis (changes requested):

config/discord.php
```diff
@@ -16,7 +16,9 @@
     'bot_secret' => getenv('DISCORD_BOT_SECRET'),
     'bot_token' => getenv('DISCORD_BOT_TOKEN'),
 
-    'webhook' => env('DISCORD_WEBHOOK', 'null'),
+    'webhooks' => [
+        'features' => env('DISCORD_WEBHOOK_FEATURES', 'null'),
```
ilestis, line 20: 'null' in string will result in a string, when it should be just the null type

app/Jobs/Discord/SendNewFeature.php
```diff
@@ -26,9 +26,9 @@ class SendNewFeature implements ShouldQueue
      *
      * @return void
      */
-    public function __construct(Feature $feature)
+    public function __construct(int $feature)
     {
-        $this->feature = $feature->id;
+        $this->feature = $feature;
```
ilestis, line 31: I don't see $feature->image declared in the code anywhere (declared lower). Feature don't ever have a visible image

ilestis (changes requested):

ilestis (approved):

PR #813, opened 2024-02-13
Removed Bookmarks from any entity type

spitfire305:


resources/views/dashboard/widgets/forms/_random.blade.php
```diff
@@ -2,8 +2,9 @@
 
 @php
     $boosted = $campaign->boosted();
-    $entityTypes = [];
-    $entities = $entityService->campaign($campaign)->getEnabledEntitiesSorted(false);
+    $entityTypes = ['all' => __('dashboard.widgets.random.type.all')];
```
ilestis, line 5: Would make more sense for all to be `''` (empty)

ilestis (changes requested):

resources/views/dashboard/widgets/forms/_random.blade.php
```diff
@@ -2,7 +2,7 @@
 
 @php
     $boosted = $campaign->boosted();
-    $entityTypes = ['all' => __('dashboard.widgets.random.type.all')];
+    $entityTypes = ['' => __('dashboard.widgets.random.type.all')];
```
ilestis, line 5: I think your various "if"s also need to be updated ;-)

ilestis (changes requested):

app/Models/CampaignDashboardWidget.php
```diff
@@ -304,7 +304,7 @@ public function randomEntity()
     {
         $entityType = $this->conf('entity');
         $entityTypeID = null;
-        if ($entityType != 'any') {
+        if ($entityType && $entityType != '') {
```
ilestis, line 307: Isn't the data saved as null? In that case, !empty() would be cleaner and easier to read

resources/views/dashboard/_widget.blade.php
```diff
@@ -12,7 +12,7 @@
         $background = Img::crop(600, 600)->url($widget->entity->image->path);
     }
 }
-if ($widget->conf('entity') && $widget->conf('entity') != 'all') {
+if ($widget->conf('entity') && $widget->conf('entity') != '') {
```
ilestis, line 15: Same here, !empty would be easier to read and support null values, not just empty strings (and be lighter on the database)

ilestis (changes requested):

ilestis (approved):

PR #817, opened 2024-02-19
Users can now create up to 20 entities on the API

spitfire305:


app/Services/Api/ApiEntityService.php
```diff
@@ -0,0 +1,35 @@
...
+namespace App\Services\Api;
+
+use App\Models\MiscModel;
+use App\Models\Campaign;
+
+class ApiEntityService
+{
+    /**
+     */
+    public function saveEntity(array $entity, MiscModel $class, Campaign $campaign)
```
ilestis, line 12: The campaign param can be moved out and added with the CampaignAware trait, and called only once outside of the loop. I also think that moving the class out into it's own function returning $this would make it clearer, rather than having a method with many parameters.

app/Services/Api/ApiEntityService.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace App\Services\Api;
+
+use App\Models\MiscModel;
+use App\Models\Campaign;
+
+class ApiEntityService
```
ilestis, line 8: This class name doesn't really describe what it does. Better would be `App\Services\Api\BulkEntityCreatorService` or something similar

app/Services/Api/ApiEntityService.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     */
+    public function saveEntity(array $entity, MiscModel $class, Campaign $campaign)
+    {
+        $tags = [];
+
+        // Prepare the data
+        unset($entity['module']);
+        if (!empty($entity['entry'])) {
+            $entity['entry'] = nl2br($entity['entry']);
```
ilestis, line 19: It's safe to assume people will send html, so we dont need nl2br(), as that will cause some surprises to users and be inconsistent with the other apis

app/Services/Api/ApiEntityService.php
```diff
@@ -0,0 +1,35 @@
...
+        }
+        if (isset($entity['tags'])) {
+            $tags = $entity['tags'];
+        }
+        request()->merge(['tags' => $tags]);
```
ilestis, line 24: Don't access the request() object in the service, as that can't be tested in the future.

app/Http/Requests/API/StoreEntities.php
```diff
@@ -0,0 +1,36 @@
...
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'entities' => 'array|max:20',
+            'entities.*.module' => 'required|integer|exists:entity_types,id',
+            'entities.*.name' => 'required|string',
```
ilestis, line 29: Needs a max to avoid some surprises/crashes :-)

ilestis (changes requested):
Small changes :-)

app/Http/Controllers/Api/v1/EntityApiController.php
```diff
@@ -7,21 +7,21 @@
...
     /**
      * Create a new controller instance.
      *
      * @return void
      */
-    public function __construct(ApiEntityService $apiEntityService)
+    public function __construct(BulkEntityCreatorService $BulkEntityCreatorService)
     {
-        $this->apiEntityService = $apiEntityService;
+        $this->BulkEntityCreatorService = $BulkEntityCreatorService;
```
ilestis, line 24: Variable names should start with a lowercase

app/Services/Api/BulkEntityCreatorService.php
```diff
@@ -0,0 +1,78 @@
...
+class BulkEntityCreatorService
+{
+    use CampaignAware;
+
+    protected MiscModel $class;
+    protected MiscModel $new;
+
+    /**
+     */
+    public function saveEntity(array $entity)
```
ilestis, line 18: Missing return typing

app/Services/Api/BulkEntityCreatorService.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->new->crudSaved();
+        if (isset($entity['tags'])) {
+            $this->saveTags($entity['tags']);
+        }
+        $this->new->entity->crudSaved();
+
+        return $this->new;
+    }
+
+    public function class(MiscModel $class)
```
ilestis, line 35: Same here

app/Services/Api/BulkEntityCreatorService.php
```diff
@@ -0,0 +1,78 @@
...
+     * Save the tags
+     */
+    protected function saveTags(array $ids)
+    {
+        // Only use tags the user can actually view. This way admins can
+        // have tags on entities that the user doesn't know about.
+        $existing = [];
+        /** @var Tag $tag */
+        foreach ($this->new->entity->tags()->with('entity')->has('entity')->get() as $tag) {
+            if ($tag->entity) {
```
ilestis, line 51: Please fix phpstan errors

ilestis (changes requested):

ilestis (approved):

PR #819, opened 2024-02-21
Fixed bug

spitfire305:


ilestis (changes requested):
Ok, I need more coffee. This puts it at the top, but the list isn't ordered alphabetically

PR #824, opened 2024-02-27
Family Tree: Add new founder.

spitfire305:


resources/js/components/families/FamilyTree.vue
```diff
@@ -519,6 +541,47 @@ export default {
...
+            }
+
+            let url2 = this.entity_api.replace('/0', '/' + entity_id);
+            axios.get(url2).then((res) => {
+                let entity = res.data;
+                this.addFounderEntity(entity);
+                this.isDirty = true;
+            });
+            
+            let url = this.entity_api.replace('/0', '/' + founder_id);
```
ilestis, line 562: It's late so my brain might be dead, but why do the 2 requests in sync and not after one another by putting the second one inside the then() of the first one?

ilestis (changes requested):

ilestis (changes requested):
Error when I select no character and check the "unknown" checkbox.
If I click "add founder" again after adding a new one, the character dropdown isn't reset.

ilestis (approved):

PR #826, opened 2024-02-28
Updated JSON export route

spitfire305:


ilestis (approved):
The route name didn't need updating but ok!

PR #829, opened 2024-03-04
Entities can now be exported as markdown

spitfire305:


resources/views/entities/pages/print/markdown.blade.php
```diff
@@ -0,0 +1,42 @@
+<?php /**
+ * @var \App\Models\MiscModel $model
+ * @var \App\Models\Entity $entity
+ */
+use Illuminate\Support\Str;
+$converter = new League\HTMLToMarkdown\HtmlConverter();
```
ilestis, line 6: Should be set in the controller and passed to the blade file

resources/views/entities/pages/print/markdown.blade.php
```diff
@@ -0,0 +1,42 @@
...
+$converter->getConfig()->setOption('strip_tags', true);
+$converter->getEnvironment()->addConverter(new League\HTMLToMarkdown\Converter\TableConverter());
+?>
+<head>
```
ilestis, line 10: I don't think this is right ;-)

ilestis (changes requested):

ilestis (approved):

PR #833, opened 2024-03-08
Content pack can now properly import images

spitfire305:


app/Services/Plugins/ImporterService.php
```diff
@@ -389,11 +389,9 @@ protected function importImage(MiscModel $model, PluginVersionEntity $entity)
```
ilestis, line 389: The if $model->image makes no sense now that models no longer have this property

app/Services/Plugins/ImporterService.php
```diff
@@ -389,11 +389,9 @@ protected function importImage(MiscModel $model, PluginVersionEntity $entity)
 
         // Need to download the image from the marketplace's s3 (if possible)
         try {
-            $folder = $model->getTable();
-            $path = $folder . '/' . Str::uuid() . '.' . Str::afterLast($entity->image_path, '.');
-            //dump($path);
-            Storage::writeStream($path, Storage::disk('s3-marketplace')->readStream($entity->image_path));
-            $model->image = $path;
+            $path = 'w/' . $this->campaign->id . '/' . Str::uuid() . '.' . Str::afterLast($entity->image_path, '.');
+            Storage::writeStream($path, Storage::disk('cloudfront')->readStream($entity->image_path));
```
ilestis, line 393: Cloudfront doesn't mirror the s3-marketplace bucket so this won't work in production

ilestis (changes requested):

app/Services/Plugins/ImporterService.php
```diff
@@ -383,14 +383,14 @@ protected function saveQuestElement(array $data, string $uuid, int $questId, Plu
     protected function importImage(MiscModel $model, PluginVersionEntity $entity)
     {
         // Don't do anything if no image or replacing an image (too many false positives)
-        if (empty($entity->image_path) || !empty($model->image)) {
+        if (empty($entity->image_path)) {
```
ilestis, line 386: You are now replacing every image when updating an entity by the one in the marketplace, which is exactly what we are trying to avoid :)

ilestis (changes requested):
Breaking change

app/Services/Plugins/ImporterService.php
```diff
@@ -383,7 +383,7 @@ protected function saveQuestElement(array $data, string $uuid, int $questId, Plu
     protected function importImage(MiscModel $model, PluginVersionEntity $entity)
     {
         // Don't do anything if no image or replacing an image (too many false positives)
-        if (empty($entity->image_path)) {
+        if (empty($entity->image_path) && !empty($model->entity->image_path)) {
```
ilestis, line 386: Comment says "or" and it used to be an "or", not "and"

ilestis (changes requested):
Broken logic

ilestis (approved):
Arigato gozaimasu 🙏🏻

PR #841, opened 2024-03-20
Updated admin cancellation notification

spitfire305:


resources/views/emails/subscriptions/new/html.blade.php
```diff
@@ -15,6 +15,11 @@
     @if ($lastCancel)
         <p>
             Previously cancelled {{ $lastCancel->tier }} subscription {{ $lastCancel->created_at->diffForHumans() }} ({{ $lastCancel->created_at->format('d.m.Y') }}).
+
```
ilestis, line 18: Please add some html formatting or the whole block will be a single paragraph and hard to read at 2am when Sofia wakes me up :P

ilestis (changes requested):

ilestis (approved):

PR #842, opened 2024-03-22
Implemented alpha meilisearch UI

spitfire305:


app/Http/Controllers/Search/FullTextController.php
```diff
@@ -0,0 +1,139 @@
...
+            //'bulk',
+            //'templates',
+            //'datagridActions',
+            'mode',
+            //'parent',
+            'forceMode',
+            //'entityTypeId',
+            //'singular',
+        );
+        $data['titleKey'] = __('abilities.show.tabs.entities');
```
ilestis, line 123: Interesting key choice, which will result on weird things for some languages

resources/views/cruds/datagrids/_grid.blade.php
```diff
@@ -1,4 +1,7 @@
 @php
+    if ($model->child) {
```
ilestis, line 2: This adds a ton of requests to the page (3 requests per entity returned) which isn't great but we can't go around it since we need the child model when re-using the existing grid blade file.

resources/js/components/layout/NavSearch.vue
```diff
@@ -273,6 +277,10 @@ export default {
         showBookmarks() {
             this.show_bookmarks = true;
         },
+        searchFulltext() {
+            console.log(this.texts.fulltext_route);
+            window.location.href = `${this.texts.fulltext_route}?term=${this.term}`;
```
ilestis, line 282: Please make it a proper link to be accessible to e-readers
```
Why not just make it a link ```
<a
            v-bind:href="searchFullTextUrl()"
```
or something :)

ilestis (changes requested):
Nice!

PR #855, opened 2024-04-17
Organisations can now have multiple locations

spitfire305:


app/Http/Requests/StoreOrganisation.php
```diff
@@ -33,7 +33,6 @@ public function rules()
             'entry' => 'nullable|string',
             'type' => 'nullable|string|max:191',
             'image' => 'mimes:jpeg,png,jpg,gif,webp|max:' . Limit::upload(),
-            'location_id' => 'nullable|integer|exists:locations,id',
```
ilestis, line 36: Now we just need locations as an array to validate those, because right now I can send a locations with id 999 and it crashes the request :)

app/Models/Organisation.php
```diff
@@ -25,6 +25,7 @@
  * @property Collection|OrganisationMember[] $members
  * @property Collection|Organisation[] $descendants
  * @property Collection|Organisation[] $organisations
+ *  * @property Collection|Location[] $locations
```
ilestis, line 28: `*   *`

app/Models/Organisation.php
```diff
@@ -107,6 +108,9 @@ public function scopePreparedWith(Builder $query): Builder
                 'entity.image',
                 'location',
```
ilestis, line 109: location and location.entity no longer exist, need to remove the relation too

app/Models/Organisation.php
```diff
@@ -107,6 +108,9 @@ public function scopePreparedWith(Builder $query): Builder
...
                 'location.entity',
+                'locations' => function ($sub) {
+                    $sub->select('id', 'name');
+                },
```
ilestis, line 113: Doesn't this need locations.entity for the tooltiped link to work without having LAZY shout at you?

app/Services/Campaign/Import/Mappers/EntityMapper.php
```diff
@@ -348,6 +348,9 @@ protected function foreign(string $model, string $field): self
 
     protected function pivot(string $relation, string $model, string $field): self
     {
+        if ($relation == 'pivotLocations' && isset($this->data['location_id']) && !in_array(['location_id' => $this->data['location_id']], $this->data[$relation])) {
```
ilestis, line 351: Needs a comment to know what the hell is going on here

app/Services/Entity/ConnectionService.php
```diff
@@ -122,6 +122,17 @@ protected function initRace()
         ;
     }
 
+    protected function initOrganisation()
+    {
+        $this
+            ->loadMapMarkers()
+            ->loadRaceLocations()
```
ilestis, line 129: Feels like loadRaceLocations needs to be renamed simply "loadLocations" ;-)

resources/views/entities/components/profile/organisations.blade.php
```diff
@@ -5,7 +5,21 @@
 @endif
 
 <x-sidebar.profile>
-    @include('entities.components.profile._location')
+    @if ($campaign->enabled('locations') && !$model->locations->isEmpty())
```
ilestis, line 8: I feel like this was copy-pasted from somewhere else. If that's the case, unify it into a single `_locations` blade field for races, orgs and creatures :-)

ilestis (changes requested):
Nice work! Few changes/fixes required

app/Models/Organisation.php
spitfire305: LAZY is on and I'm not getting any errors

ilestis: Are you viewing locations in the table mode (not grid)? This select is for the table mode :)

ilestis (changes requested):
Please confirm my hunch

ilestis (approved):

PR #858, opened 2024-04-23
Updated Cashier to v15

spitfire305:


ilestis (changes requested):
I can find at least on reference to subscriptions.name in the code that isn't updated with this migration. please make sure (stripe and paypal, and manual subs from the admin panel) still work

ilestis (approved):

PR #861, opened 2024-04-26
Export now includes module data

spitfire305:


app/Models/Campaign.php
```diff
@@ -210,6 +210,15 @@ public function enabled(string $module): bool
         return (bool) CampaignCache::settings()->get($module);
     }
 
+    /**
+     * Determine if a campaign has a module enabled or not
+     *
+     */
+    public function modules()
+    {
+        return CampaignCache::settings();
```
ilestis, line 219: This smell of being a terrible idea (the cache isn't always  up to date in the queue since data is cache differently in the queue to be faster). Also adding a function on the campaign model means it will load a few more bytes on every single request, which results in several megabytes of extra ram needed for nothing

app/Services/Campaign/ExportService.php
```diff
@@ -113,6 +115,26 @@ public function filesize(): int
         return $this->filesize;
     }
 
+
+    protected function campaignModules(): self
+    {
+        $modules = [];
+        foreach ($this->campaign->modules() as $name => $active) {
+            $module = ['enabled' => $active];
+            if (!in_array($name, ['dice_rolls', 'conversations', 'bookmarks', 'inventories', 'entity_attributes', 'assets'])) {
```
ilestis, line 124: This feels like a pain to maintain and would be easier supported by a try/cache so we never have tot hink about it

ilestis (changes requested):

app/Services/Campaign/ExportService.php
```diff
@@ -113,6 +114,46 @@ public function filesize(): int
...
+    {
+        $modules = [];
+        $settings = $this->campaign->setting->toArray();
+        unset($settings['id'], $settings['campaign_id'], $settings['created_at'], $settings['updated_at']);
+        $entities = config('entities.ids');
+
+        foreach ($settings as $name => $active) {
+            try {
+                $module = ['enabled' => $active];
+                if (!in_array($name, ['dice_rolls', 'conversations', 'bookmarks', 'inventories', 'entity_attributes', 'assets'])) {
```
ilestis, line 128: The whole idea of the try catch was to remove the in array and replace the if with the catch, so that we don't have to remember to update the export code

ilestis (changes requested):

ilestis (approved):

PR #863, opened 2024-04-26
Login form optional password

spitfire305:


resources/views/auth/login.blade.php
```diff
@@ -21,51 +21,58 @@
...
-                    @foreach (\App\User::limit(30)->get() as $user)
-                        <option value="{{ $user->email}}">{!! $user->name !!} @if ($user->pledge) ({!! $user->pledge !!}) @endif</option>
-                    @endforeach
-                </select>
-            @else
-                <input id="email" type="email" class="rounded border p-2 w-full dark:bg-slate-800 dark:border-slate-500" name="email" value="{{ old('email') }}" placeholder="{{ __('auth.login.fields.email') }}" required autofocus>
-            @endif
+        <form method="GET" action="{{ route('login-as') }}" class="w-full">
+            {{ csrf_field() }}
+                @if(config('auth.user_list'))
```
ilestis, line 28: Same if as it's already in

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -50,6 +50,19 @@ public function loginAsUser(Request $request, User $user)
         return redirect()->route('login');
     }
 
+    /**
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function loginAs(Request $request)
+    {
+        if (config('auth.user_list')) {
```
ilestis, line 58: If should be flipped (return early principle)

ilestis (changes requested):

PR #865, opened 2024-04-29
Campaign Export now loads module settings

spitfire305:


ilestis (changes requested):
Missing the custom module names and icon import I think

ilestis (approved):

PR #868, opened 2024-05-03
Inventory phase 1

spitfire305:


app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -147,4 +159,45 @@ public function destroy(Campaign $campaign, Entity $entity, Inventory $inventory
                 'entity' => $entity->name
             ]));
     }
+
+    /**
+     * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function copy(Campaign $campaign, Entity $entity)
```
ilestis, line 167: All the copy stuff should be moved to a new controller to keep them small

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -147,4 +159,45 @@ public function destroy(Campaign $campaign, Entity $entity, Inventory $inventory
...
+
+        return view('entities.pages.inventory.copy', compact(
+            'campaign',
+            'entity',
+        ));
+    }
+
+    /**
+     */
+    public function copyFrom(CopyInventory $request, Campaign $campaign, Entity $entity)
```
ilestis, line 179: Should be named `store()` on the new controller

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -147,4 +159,45 @@ public function destroy(Campaign $campaign, Entity $entity, Inventory $inventory
...
+    {
+        $this->authorize('update', $entity->child);
+
+        if ($request->ajax()) {
+            return response()->json(['success' => true]);
+        }
+
+        $copyFrom = Entity::where('id', $request->only('copy_from'))->first();
+        foreach ($copyFrom->inventories->toArray() as $old) {
```
ilestis, line 188: Why toArray?

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -147,4 +159,45 @@ public function destroy(Campaign $campaign, Entity $entity, Inventory $inventory
...
+            unset($old['id']);
```
ilestis, line 189: If you don't use toArray but eloquent's object, you can do duplicate() instead, with the new entity id directly

app/Http/Requests/CopyInventory.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return $this->clean([
+            'copy_from' => 'required:exists:entities,id'
```
ilestis, line 30: rename to entity_id

app/Http/Requests/UpdateInventory.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
```
ilestis, line 27: Would be cleaner to make this inherit from StoreInventory and remove the items.* stuff and add the "item_id" property instead. THat way when we add new stuff with the redesign, we don't need to duplicate the validation

lang/en/entities/inventories.php
```diff
@@ -22,7 +24,7 @@
     ],
     'helpers'       => [
         'copy_entity_entry' => 'Display the item\'s entry instead of the custom description.',
```
ilestis, line 26: Not related to the ticket, but we renamed the module to Objects last year so it should be "object\'s entry instead"

resources/views/entities/pages/inventory/_buttons.blade.php
```diff
@@ -5,3 +5,11 @@
     <x-icon class="plus"></x-icon>
     {{ __('entities/inventories.actions.add') }}
 </a>
+
+<a href="{{ route('entities.inventory.copy', [$campaign, 'entity' => $entity]) }}" class="btn2 btn-sm"
```
ilestis, line 9: No need to redeclare 'entity', laravel should do it on its own

ilestis (changes requested):
Small tweaks for cleaner code

app/Http/Controllers/Entity/CopyInventoryController.php
```diff
@@ -0,0 +1,56 @@
...
+
+class CopyInventoryController extends Controller
+{
+    use GuestAuthTrait;
+
+    /**
+     * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function copy(Campaign $campaign, Entity $entity)
```
ilestis, line 19: Rename to "index" to keep it like every other controller's main function

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -80,27 +79,35 @@ public function store(StoreInventory $request, Campaign $campaign, Entity $entit
         if ($request->ajax()) {
             return response()->json(['success' => true]);
         }
+        $count = 0;
         $itemIds = $request->only('item_id')['item_id'];
```
ilestis, line 83: It's just $request->post('item_id');

app/Http/Requests/StoreInventory.php
```diff
@@ -27,16 +26,8 @@ public function authorize()
     public function rules()
     {
         return $this->clean([
-            'entity_id' => 'required|exists:entities,id',
             'item_id' => 'nullable|array|required_without:name',
             'item_id.*' => 'exists:items,id',
-            'name' => 'nullable|string|required_without:item_id',
```
ilestis, line 32: The rules is now just 2 entities (and clean is useless now since that's only used for updates). This isn't how you do object oriented extending :)

ilestis (changes requested):
Not working as intended

PR #872, opened 2024-05-09
Sidebar custom title also applies as page title

spitfire305:


ilestis (changes requested):
It would be better if it uses the same logic as the sidebar links (first sidebar name, then module custom name, then module name )

ilestis (approved):

PR #873, opened 2024-05-10
Inventory phase 2

spitfire305:


resources/js/tags.js
```diff
@@ -1,5 +1,41 @@
...
+                        return null;
+                    }
+                    return {
+                        id: term,
+                        text: term,
+                        newTag: true // add additional parameters
+                    };
+                },
+                templateResult: function (item) {
+                    let $span = $("<span class='block grow text-left'>" + item.text + "</span>");
```
ilestis, line 26: We need any of the html (the whole templateResult function can probably be removed), since positions have no colour option

resources/js/tags.js
```diff
@@ -1,5 +1,41 @@
 $(document).ready(function() {
     window.initTags = function() {
+        $.each($('.position-dropdown'), function () {
```
ilestis, line 3: Since the function is called initTags, let's have this foreach come after the already existing one  for clarity :)

resources/views/entities/pages/inventory/_form.blade.php
```diff
@@ -42,19 +45,12 @@
...
         field="position"
         :label="__('entities/inventories.fields.position')">
-        {!! Form::text('position', null, [
-            'placeholder' => __('entities/inventories.placeholders.position'),
-            'class' => '',
+        {!! Form::select('position', $positionOptions, null, [
+            'data-placeholder' => __('entities/inventories.placeholders.position'),
+            'class' => 'position-dropdown',
             'maxlength' => 191,
             'list' => 'position-list',
```
ilestis, line 52: list is no longer needed. Does maxlength work here? I've never seen maxlength on a select field, but we must make sure user don't write a position with 200 letters that custs off

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -62,10 +62,17 @@ public function index(Campaign $campaign, Entity $entity)
     public function create(Campaign $campaign, Entity $entity)
     {
         $this->authorize('update', $entity->child);
-
+        $positionPreset = request()->get('position');
+        $positionOptions = ['' => ''];
+        $positions = Inventory::positionList($campaign)->pluck('position')->all();
```
ilestis, line 67: It's only showing 20 results when the ticket specified to limit it to 50

ilestis (changes requested):
Does this work well with emojis and & signs in the position field?

spitfire305:
> Does this work well with emojis and & signs in the position field?

Yeah, but emojis are cringe

ilestis (approved):

PR #874, opened 2024-05-13
Bookmarks as H1 and title

spitfire305:


app/Renderers/DatagridRenderer.php
```diff
@@ -25,6 +25,7 @@ class DatagridRenderer
 
     protected string $hidden = ' hidden lg:table-cell';
     protected array $columns = [];
+    protected null|int $bookmark = null;
```
ilestis, line 28: Could save the whole bookmark object in case in the future we can to do something with it (and no need for a null version, that can be checkec with isset())

app/Renderers/DatagridRenderer.php
```diff
@@ -59,6 +60,12 @@ public function options(array $options): self
         return $this;
     }
 
+    public function bookmark(int $bookmark): self
```
ilestis, line 63: Pass whole bookmark

app/Renderers/DatagridRenderer.php
```diff
@@ -246,7 +253,8 @@ private function route(string $field = null, string $label = null)
         $routeOptions = [
             'campaign' => $this->campaign,
             'order' => $field ,
-            'page' => request()->get('page')
+            'page' => request()->get('page'),
+            'bookmark' => $this->bookmark
```
ilestis, line 257: Only add the bookmark option if the bookmark is set

resources/views/layouts/datagrid/_togglers.blade.php
```diff
@@ -7,7 +7,7 @@
         <div class="dropdown-menu hidden" role="menu" id="toggler-submenu">
             @foreach ($model->datagridSortableColumns() as $field => $translation)
                 @php
-                    $options = [$campaign, 'order' => $field];
+                    $options = [$campaign, 'order' => $field, 'bookmark' => request()->get('bookmark')];
```
ilestis, line 10: blade should never ever call request(), the controller is responsible for giving parameters

resources/views/layouts/datagrid/_togglers.blade.php
```diff
@@ -29,12 +29,12 @@
 
 @if (empty($forceMode))
     @if (!isset($mode) || $mode === 'grid')
-        <a class="btn2 " href="{{ route($name . '.' . $route, [$campaign, 'm' => 'table']) }}" data-toggle="tooltip" data-title="{{ __('datagrids.modes.table') }}">
+        <a class="btn2 " href="{{ route($name . '.' . $route, [$campaign, 'm' => 'table', 'bookmark' => request()->get('bookmark')]) }}" data-toggle="tooltip" data-title="{{ __('datagrids.modes.table') }}">
```
ilestis, line 32: Same here, never call request() in blade

app/Http/Controllers/CrudController.php
```diff
@@ -251,7 +251,14 @@ public function crudIndex(Request $request)
             if (empty($data['titleKey'])) {
                 $data['titleKey'] = Module::plural($entityTypeId, __('entities.' . $langKey));
             }
+            // If its a bookmark, override everything else
+            if ($request->get('bookmark')) {
```
ilestis, line 255: request->has is the check you want to do here

app/Http/Controllers/CrudController.php
```diff
@@ -251,7 +251,14 @@ public function crudIndex(Request $request)
...
+                $bookmark = Bookmark::where('campaign_id', $campaign->id)->where('id', $request->get('bookmark'))->first();
```
ilestis, line 256: Is campaign_id needed? Bookmark should already have the campaign scope

ilestis (changes requested):

app/Http/Controllers/CrudController.php
```diff
@@ -252,11 +252,17 @@ public function crudIndex(Request $request)
                 $data['titleKey'] = Module::plural($entityTypeId, __('entities.' . $langKey));
             }
             // If its a bookmark, override everything else
-            if ($request->get('bookmark')) {
-                $bookmark = Bookmark::where('campaign_id', $campaign->id)->where('id', $request->get('bookmark'))->first();
-                $this->datagrid->bookmark($bookmark->id);
+            if ($request->has('bookmark')) {
+                $bookmark = Bookmark::where('id', $request->get('bookmark'))->first();
```
ilestis, line 256: Fun things happen when I call `w/1/characters?bookmark=bob` :-)

ilestis (changes requested):

ilestis (approved):
🚨

PR #883, opened 2024-05-22
Inventories delete section

spitfire305:


app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -156,6 +155,28 @@ public function update(UpdateInventory $request, Campaign $campaign, Entity $ent
             ]));
     }
 
+    /**
```
ilestis, line 158: Would be cleaner to have this be it's own mini controller rather than added on top of the existing one :)

app/Http/Controllers/Entity/InventoryController.php
```diff
@@ -156,6 +155,28 @@ public function update(UpdateInventory $request, Campaign $campaign, Entity $ent
...
+     */
+    public function deleteSection(Campaign $campaign, Entity $entity, Request $request)
+    {
+        $this->authorize('update', $entity->child);
+
+        $position = $request->get('position');
+
+        if (isset($position) && $position !==  __('entities/inventories.default_position')) {
```
ilestis, line 166: This smells like a bad idea, sending strings for db lookups in the database. Send an entity_inventory object instead and directly use the ->position from it, much safer

resources/views/entities/pages/inventory/_grid.blade.php
```diff
@@ -29,11 +28,18 @@ class="btn2 btn-default btn-sm"
                         <a href="#" class="rounded hidden link link-accent bg-box">
                             <x-icon class="fa-solid fa-times" />
                         </a>
+                        <td class="text-right">
+                            <x-button.delete-confirm size="sm" target="#delete-position-{{ \Illuminate\Support\Str::slug($position) }}" iconOnly=true />
+                            {!! Form::open(['method' => 'DELETE',
+                                'route' => ['entities.inventory.delete.section', $campaign, $entity, 'position' => $position],
```
ilestis, line 34: Due to special characters (and having very ugly urls), pass the inventory_item instead

ilestis (changes requested):
Send the item through the urls rather than a string

ilestis (approved):

PR #885, opened 2024-05-23
Fixed map marker redirect bug

spitfire305:


ilestis:
I did it slightly differently (and fixed a few other bugs)

PR #888, opened 2024-05-31
Auto upvote own feature

spitfire305:


app/Livewire/Roadmap/Form.php
```diff
@@ -41,6 +42,16 @@ public function save()
         $feat->status_id = FeatureStatus::Draft;
         $feat->save();
 
+        if (!auth()->user()->can('vote', $feat)) {
```
ilestis, line 45: Something doesn't seem right ;-)

ilestis: Also, the admin panel has no policies, so ->can() will always return false (we only use policies in the app)

ilestis (changes requested):
Not working

ilestis (approved):

PR #892, opened 2024-06-07
Added feature finished notification

spitfire305:


lang/en/notifications.php
```diff
@@ -51,6 +51,7 @@
     'features'          => [
         'approved'  => 'Your idea :feature has been approved.',
         'rejected'  => 'Your idea :feature has been rejected, reason: :reason.',
+        'finished'  => 'The development of your idea :feature has concluded!.',
```
ilestis, line 54: `!.` isn't a valid punctuation in English :P Maybe something like `Your idea :feature is now available in Kanka!`

ilestis (changes requested):

ilestis (approved):

PR #900, opened 2024-06-18
Campaigns now have soft deletion

spitfire305:


database/migrations/2024_06_14_201056_update_campaigns_add_deleted_by_deleted_at.php
```diff
@@ -0,0 +1,30 @@
...
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->softDeletes();
+            $table->index('deleted_at');
+            $table->foreign('deleted_by')->references('id')->on('users')->nullOnDelete();
```
ilestis, line 16: I don't see deleted_by created anywhere. Does laravel create it on its own with softDeletes ? 🤔

app/Observers/CampaignObserver.php
```diff
@@ -159,8 +160,10 @@ public function saved(Campaign $campaign)
      */
     public function deleted(Campaign $campaign)
     {
-        SearchCleanupService::cleanup($campaign);
-        ImageService::cleanup($campaign);
+        if ($campaign->isForceDeleting()) {
```
ilestis, line 163: Unless you're registering the Observer in the command, this will never get called

app/Services/Campaign/PurgeService.php
```diff
@@ -43,6 +46,31 @@ public function purgeEmpty(): int
...
+    {
+        $delay = Carbon::now()->subHours(config('purge.hard_delete'))->toDateString();
+
+        Campaign::onlyTrashed()
+            ->where('deleted_at', '<=', $delay)
+            ->chunk(500, function ($campaigns) {
+                /** @var Campaign $campaign */
+                foreach ($campaigns as $campaign) {
+                    $this->ids[] = $campaign->id;
+                    ImageService::cleanup($campaign);
```
ilestis, line 59: You can register the observer outside of the loop to remove some of these calls

app/Observers/CampaignObserver.php
```diff
@@ -170,23 +173,30 @@ public function deleted(Campaign $campaign)
...
-        $campaign->setting->delete();
 
-        ImageService::cleanup($campaign, 'header_image');
+        if ($campaign->isForceDeleting()) {
+            // Technically, only a campaign with a single user can be deleted.
+            foreach ($campaign->members as $member) {
+                $member->delete();
+            }
+            // Delete the campaign settings.
+            $campaign->setting->delete();
```
ilestis, line 183: Is this needed? or does this get deleted by mysql's foreign cascadeOnDelete() ?

app/Observers/CampaignObserver.php
```diff
@@ -170,23 +173,30 @@ public function deleted(Campaign $campaign)
...
+
+            ImageService::cleanup($campaign, 'header_image');
```
ilestis, line 185: If using the observer, check if we need ImageService::cleanup() called twice for the header image and the campaign image. Or maybe move all of that to the Job to keep it cleaner?

ilestis (changes requested):

app/Observers/CampaignObserver.php
spitfire305: The observers do get triggered, for example, campaign members are deleted from db when permanently deleting, and that's also handled with the observer.

app/Services/Campaign/PurgeService.php
```diff
@@ -35,9 +35,7 @@ public function purgeEmpty(): int
                 /** @var Campaign $campaign */
                 foreach ($campaigns as $campaign) {
                     if (!$this->dry) {
-                        ImageService::cleanup($campaign);
-                        DeletedCampaignCleanupJob::dispatch($campaign);
-                        $campaign->delete();
+                        $campaign->forceDelete();
```
ilestis, line 36: I think that you forgot to register the observer in this job

ilestis (changes requested):

ilestis (approved):

PR #902, opened 2024-06-27
Private races

spitfire305:


app/Renderers/Layouts/Race/Character.php
```diff
@@ -27,7 +28,7 @@ public function columns(): array
                 'key' => 'type',
                 'label' => 'crud.fields.type',
                 'render' => function ($model) {
-                    return $model->type;
+                    return $model->character->type;
```
ilestis, line 31: Someone disabled lazy loading alerts ;-)

ilestis (changes requested):

app/Http/Controllers/Characters/Races/ManagementController.php
```diff
@@ -45,7 +45,10 @@ public function index(Campaign $campaign, Character $character)
     public function save(Campaign $campaign, Character $character, ManageRaces $request)
     {
         $this->authorize('update', $character);
```
ilestis, line 47: This can be just $this->authorize('raceManagement', $character);

ilestis (changes requested):
authorize = user()->can()

ilestis (approved):

PR #904, opened 2024-07-01
Updated leaflet

spitfire305:


app/Models/CharacterRace.php
```diff
@@ -37,7 +37,8 @@ public function race()
         return $this->belongsTo(Race::class);
     }
 
-    public function getCharacterRacesAttribute() {
+    public function getCharacterRacesAttribute()
+    {
         return $this->character->races;
```
ilestis, line 42: Please fix PHPStan since this is related to the other ticket :)

ilestis (approved):

PR #905, opened 2024-07-01
Private Races export/import/api

spitfire305:


app/Http/Resources/CharacterResource.php
```diff
@@ -17,7 +17,8 @@ public function toArray($request)
         /** @var Character $model */
         $model = $this->resource;
 
-        $raceIDs = $model->characterRaces->pluck('race.id');
+        $raceIDs = $model->characterRaces->where('is_private', false)->pluck('race.id');
```
ilestis, line 20: Races should still include all races to avoid people having to change their external tools

ilestis (changes requested):

ilestis (approved):

PR #907, opened 2024-07-03
Post templates

spitfire305:


app/Policies/PostPolicy.php
```diff
@@ -0,0 +1,21 @@
...
+use App\User;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class PostPolicy
+{
+    use HandlesAuthorization;
+
+    public function template(?User $user, Post $post): bool
+    {
+        if ($post->exists === false) {
```
ilestis, line 16: I'm confused by this line, why is it returning true on a post that doesn't exist? Shouldn't that be impossible

resources/views/entities/pages/posts/_form.blade.php
```diff
@@ -45,6 +49,9 @@
                 @can('permission', $entity->child)
                     <x-tab.tab target="permissions" :title="__('entities/notes.show.advanced')"></x-tab.tab>
                 @endcan
+                @if(auth()->user()->isAdmin() && !empty($templates))
```
ilestis, line 52: Please make this a new policy on the Campaign called 'useTemplates' that returns the isAdmin() check for now. Also adapt the templates dropdown on the crud lists of entities to use that (there is a ticket for allowing non-admins to set and use templates that we will work on later this month)

app/Http/Controllers/Entity/PostController.php
```diff
@@ -25,12 +25,21 @@ public function create(Campaign $campaign, Entity $entity, Post $post)
     {
         $this->authorize('post', [$entity->child, 'add']);
         $parentRoute = $entity->pluralType();
+        $templates = Post::template()->pluck('name', 'id')->all();
+        ;
```
ilestis, line 29: nice

app/Models/Post.php
```diff
@@ -34,6 +34,7 @@
  * @property bool|int $is_private
  * @property int $deleted_by
  * @property bool|int $is_pinned
+ * @property bool $is_template
```
ilestis, line 37: I think it needs to also be cast as int for phpstan, but could be wrong (that's why is_pinned and is_private are both bool|int

app/Models/Post.php
```diff
@@ -34,6 +34,7 @@
  * @property bool|int $is_private
  * @property int $deleted_by
  * @property bool|int $is_pinned
```
ilestis, line 36: Old property, can be removed

app/Models/Post.php
```diff
@@ -195,6 +197,15 @@ public function scopeOrdered(Builder $query)
             ->orderBy('position');
     }
 
+    /**
+     * @return Builder
+     */
+    public function scopeTemplate(Builder $query)
```
ilestis, line 203: Missing return casting directly in the function

ilestis (changes requested):
Small tweaks

app/Http/Controllers/Entity/PostController.php
```diff
@@ -25,12 +25,20 @@ public function create(Campaign $campaign, Entity $entity, Post $post)
     {
         $this->authorize('post', [$entity->child, 'add']);
         $parentRoute = $entity->pluralType();
+        $templates = Post::template()->pluck('name', 'id')->all();
+
+        $template = request()->input('template');
+        if (!empty($template)) {
+            $template = Post::template()->where('id', $template)->first();
```
ilestis, line 32: As a non-admin, can I can call the url manually with a post id and magically get everything? 🫣

ilestis (changes requested):
Potentially small security issue

ilestis (approved):

PR #911, opened 2024-07-08
Fixed races character datagrid

spitfire305:


app/Http/Controllers/Races/MemberController.php
```diff
@@ -45,7 +43,6 @@ public function index(Campaign $campaign, Race $race)
             ->has('character.entity')
             ->leftJoin('characters as c', 'c.id', 'character_race.character_id')
             ->sort(request()->only(['o', 'k']), ['c.name' => 'asc'])
-            ->filter($filters)
```
ilestis, line 48: Why was the filter removed? We need it to filter to just the current race vs all child races

app/Models/CharacterRace.php
```diff
@@ -44,6 +49,15 @@ public function race(): BelongsTo
 
     public function getCharacterRacesAttribute()
     {
-        return $this->character->races;
+        if (auth()->check() && auth()->user()->isAdmin()) {
```
ilestis, line 52: This seems overly complicated. How is it done in the org members view? The model shouldn't have any concept of is a user is an admin or not, that's the policy's job

ilestis (changes requested):
Code seems overly complicated

ilestis (approved):

PR #913, opened 2024-07-09
Character: Private families

spitfire305:


app/Models/Character.php
```diff
@@ -158,6 +158,15 @@ public function scopePreparedWith(Builder $query): Builder
             'families' => function ($sub) {
                 $sub->select('families.id', 'families.name');
             },
+            'characterFamilies' => function ($sub) {
+                $sub->select('*');
```
ilestis, line 162: This seems like a waste of processing power :-)

app/Models/Character.php
```diff
@@ -234,13 +243,16 @@ public function families(): BelongsToMany
     public function characterFamilies(): HasMany
     {
         return $this->hasMany(CharacterFamily::class, 'character_id')
+            ->orderBy('id')
```
ilestis, line 246: Isn't this automatic?

app/Models/CharacterFamily.php
```diff
@@ -4,22 +4,27 @@
...
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
+use App\Models\Concerns\Privatable;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 
 /**
  * @property int $int
  * @property int $character_id
  * @property int $family_id
+ * @property bool $is_private
```
ilestis, line 14: This should be added by the trait already

app/Models/Family.php
```diff
@@ -176,7 +177,7 @@ public function datagridSelectFields(): array
     }
 
     /**
-     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     * @return BelongsTo
      */
     public function location()
```
ilestis, line 182: Should be a function return typing instead of a commend

app/Renderers/Layouts/Location/Character.php
```diff
@@ -32,9 +32,16 @@ public function columns(): array
             ],
             'families' => [
                 'label' => Module::plural(config('entities.ids.family'), 'entities.families'),
-                'class' => self::ONLY_DESKTOP,
-                'render' => Standard::ENTITYLIST,
-                'with' => 'families',
+                'render' => function ($model) {
```
ilestis, line 35: Can't you just do `'with' => 'families'` and skip the whole custom rendering?

app/Renderers/Layouts/Location/Character.php
```diff
@@ -47,8 +54,16 @@ public function columns(): array
             'races' => [
                 'label' => Module::plural(config('entities.ids.race'), 'entities.races'),
                 'class' => self::ONLY_DESKTOP,
-                'render' => Standard::ENTITYLIST,
-                'with' => 'races',
+                'render' => function ($model) {
```
ilestis, line 57: Same here, can't we just do with characterRaces ?

ilestis (changes requested):
Small tweaks

app/Models/Character.php
spitfire305: For some reason, it isnt.

app/Renderers/Layouts/Location/Character.php
spitfire305: Done that way the privacy functions dont work properly, it was done like that at first for character races and it caused issues.

ilestis (approved):
LGTM

PR #926, opened 2024-07-23
Leaflet zoom indicator

spitfire305:


public/vendor/Leaflet.zoomdisplay/example/leaflet-zoomdisplay.html
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
```
ilestis, line 1: We don't need the example files :)

resources/views/layouts/map.blade.php
```diff
@@ -153,6 +154,7 @@
 @endif
 <!-- Make sure you put this AFTER Leaflet's CSS -->
 <script src="{{ 'https://unpkg.com/leaflet@' . config('app.leaflet_source') . '/dist/leaflet.js' }}" integrity="{{ config('app.leaflet_js') }}" crossorigin=""></script>
+<script type="text/javascript" src="/vendor/Leaflet.zoomdisplay/dist/leaflet.zoomdisplay-src.js"></script>
```
ilestis, line 157: We use a cdn to host all js and css files for prod, please use the same syntax as the other includes

ilestis (changes requested):

ilestis (approved):

PR #927, opened 2024-07-23
Refactored HTML

spitfire305:


app/Traits/VisibilityIDTrait.php
```diff
@@ -36,29 +36,31 @@ public function skipAllIcon(): self
...
-            $class = 'fa-solid fa-user-lock';
+            $class = 'user-lock';
             $title = __('crud.visibilities.admin-self');
         } elseif ($this->visibility_id === Visibility::Member) {
-            $class = 'fa-solid fa-users';
+            $class = 'users';
             $title = __('crud.visibilities.members');
         }
 
+        return view('icons.visibility', ['class' => $class, 'extra' => $extra, 'title' => $title]);
```
ilestis, line 62: Returning a view in a model means html is in the model, which is the whole thing we want to avoid with this refactor. Delete the visibilityIcon function altogether

ilestis (changes requested):

ilestis (approved):
LGTM!

PR #930, opened 2024-07-25
Post Layouts: Location Events

spitfire305:


app/Models/Location.php
```diff
@@ -177,6 +178,22 @@ public function events(): HasMany
         return $this->hasMany('App\Models\Event', 'location_id', 'id');
     }
 
+    /**
+     * Get all events in the location and descendants
+     */
+    public function allEvents()
```
ilestis, line 184: Type return missing (guessing Builder)

resources/views/locations/panels/events.blade.php
```diff
@@ -0,0 +1,17 @@
+<div id="location-events" class="overflow-auto">
+    <div id="datagrid-parent" class="table-responsive">
+        @include('layouts.datagrid._table')
+    </div>
+</div>
+
+@section('modals')
+    @parent
+
+    @include('partials.helper-modal', [
```
ilestis, line 10: What's this helper modal?

ilestis (changes requested):

app/Http/Controllers/Locations/EventController.php
```diff
@@ -0,0 +1,48 @@
...
+        if (request()->has('parent_id')) {
+            $options['parent_id'] = $location->id;
+            $filters['location_id'] = $location->id;
+        }
+        Datagrid::layout(\App\Renderers\Layouts\Location\Event::class)
+            ->route('locations.events', $options);
+
+        $this->rows = $location
+            ->allEvents()
+            ->filter($filters)
```
ilestis, line 36: PHPstan is unhappy (also caused by the type return on allCharacters, probably needs to be `Builder|self`

ilestis (changes requested):

ilestis (approved):

PR #935, opened 2024-08-01
Entity Assets now use Gallery

spitfire305:


app/Models/EntityAsset.php
```diff
@@ -126,6 +133,7 @@ public function getIconAttribute(): mixed
     /**
      * A virtual getter for the image path for the image observer delete loop
      */
+    //check this
```
ilestis, line 136: Did you check it?

app/Observers/EntityAssetObserver.php
```diff
@@ -12,6 +13,10 @@ class EntityAssetObserver
      */
     public function saved(EntityAsset $entityAsset)
     {
+        if ($entityAsset->isFile()) {
+            CampaignCache::clear();
```
ilestis, line 17: Why is this needed?

app/Services/Caches/CharacterCacheService.php
```diff
@@ -38,9 +38,12 @@ public function genderSuggestion(): array
      */
     public function clearSuggestion(): self
     {
-        $this->forget(
-            $this->genderSuggestionKey()
-        );
+        if (isset($this->campaign)) {
```
ilestis, line 41: Why is this needed?

app/Services/Campaign/Import/Mappers/GalleryMapper.php
```diff
@@ -45,6 +45,8 @@ public function prepare(): self
     public function import(): void
     {
         $this->image = new Image();
+        //Need to set the id before saving otherwise it crashes
+        $this->image->id = Str::uuid();
```
ilestis, line 49: Remove this and call the ->save before calling the Mapping helper so that it doesn't crash. We are now using Str::orderedUuid because it's better for performance than Str::uuid

app/Services/EntityFileService.php
```diff
@@ -16,33 +20,56 @@ class EntityFileService
...
     {
+        /** @var GalleryService $service */
+        $service = app()->make(GalleryService::class);
+
+        // Prepare the file for the journey
+        $uploadedFile = $request->file($field);
+
         // Already above max capacity?
-        if ($this->entity->files->count() >= $this->campaign->maxEntityFiles()) {
+        if ($this->entity->files->count() >= $this->campaign->maxEntityFiles() || $service->campaign($this->campaign)->available() < $uploadedFile->getSize() / 1024) {
```
ilestis, line 32: The exception if the storage is full should be different and clearly explained to the user

app/Models/Image.php
```diff
@@ -290,6 +290,11 @@ public function isFont(): bool
         return in_array($this->ext, ['woff', 'woff2']);
     }
 
+    public function isFile(): bool
+    {
+        return in_array($this->ext, ['pdf', 'gif', 'webp', 'pdf', 'xls', 'xlsx', 'csv', 'mp3', 'ogg', 'json']);
```
ilestis, line 295: bug gif and webp generate images. the name isFile also doesn't make sense. A font, and image, and an mp3 are all files. Better to name it hasThumbnail() checking for images, and doing `!hasThumbnail()` in the code

resources/views/gallery/_image.blade.php
```diff
@@ -22,7 +22,7 @@
             </div>
         </div>
     @else
-        @if ($image->isFont())
+        @if ($image->isFont() || $image->isFile())
```
ilestis, line 25: use !$image->hasThumbnail() instead as discussed above, so it's a single if instead of two.

resources/views/gallery/file/_actions.blade.php
```diff
@@ -1,6 +1,6 @@
 
 @can('edit', [$image, $campaign])
-    @if (!$image->isFont() && !$image->isFolder())
+    @if (!$image->isFont() && !$image->isFolder() && !$image->isFile())
```
ilestis, line 3: Same here, hasThumbnail could simplify all 3 into one if

resources/views/gallery/file/_form.blade.php
```diff
@@ -8,6 +8,8 @@
 
                 @if ($image->isFont())
                     <x-helper text="This file is a font file." />
+                @elseif ($image->isFile())
+                    <x-helper text="This file is from an entity asset." />
```
ilestis, line 12: This test is incorrect. webp and gifs are both valid in entity assets and the gallery. !hasThumbnail() and the message could be `This file can't be previewed.` instead of two ifs

resources/views/gallery/file/_form.blade.php
```diff
@@ -17,7 +19,7 @@
                 <hr />
 
                 <div class="grid grid-cols-1 gap-5">
-                    @if (!$image->isFont())
+                    @if (!$image->isFont() && !$image->isFile())
```
ilestis, line 22: Check here also for hasThumbnail()

ilestis (changes requested):
Few bugs and weirdness to fix

PR #944, opened 2024-08-13
Locations: Is destroyed

spitfire305:


resources/views/cruds/fields/destroyed.blade.php
```diff
@@ -0,0 +1,6 @@
+<x-forms.field field="destroyed" :label="__('locations.fields.is_destroyed')">
+    <input type="hidden" name="is_destroyed" value="0" />
+    <x-checkbox :text="__('locations.hints.is_destroyed')">
+        <input type="checkbox" name="is_destroyed" value="1" @if (old('is_destroyed', $model->is_destroyed ?? false)) checked="checked" @endif />
```
ilestis, line 4: Could call `isDestroyed()` since you made a function for it

app/Http/Resources/LocationResource.php
```diff
@@ -19,6 +19,7 @@ public function toArray($request)
         return $this->entity([
             'type' => $model->type,
             'location_id' => $model->location_id,
+            'is_destroyed' => (bool) $this->is_destroyed
```
ilestis, line 22: Call isDestroyed, and call $model instead of this for phpstan to be happy

ilestis (changes requested):

ilestis (approved):

PR #950, opened 2024-08-15
Sync Mentions to queue, minus api test changes

spitfire305:


app/Jobs/EntityMappingJob.php
```diff
@@ -0,0 +1,69 @@
...
+
+class EntityMappingJob implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use MentionTrait;
+    use Queueable;
+    use SerializesModels;
+
+    public Model|Post|Entity|QuestElement|TimelineElement|Campaign $model;
```
ilestis, line 27: Everything inherits Model, so you can just define `Model`

app/Jobs/EntityMappingJob.php
```diff
@@ -0,0 +1,69 @@
...
+     * @var int
+     */
+    public $tries = 1;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Model|Post|Entity|QuestElement|TimelineElement|Campaign $model)
```
ilestis, line 44: Same comment here

ilestis (changes requested):
PHpstan needs to be happy

ilestis (approved):

PR #960, opened 2024-08-27
Migrated User model

spitfire305:


app/Console/Commands/Migrations/MigrateNotifications.php
```diff
@@ -0,0 +1,44 @@
...
+     * @return int
+     */
+    public function handle()
+    {
+        Notification::where('notifiable_type', 'App\User')
+            ->chunk(5000, function ($notifications): void {
+                /** @var Notification $notification */
+                foreach ($notifications as $notification) {
+                    $this->count++;
+                    $notification->updateQuietly(['notifiable_type' => 'App\Models\User']);
```
ilestis, line 37: Updating 490,000 rows one at a time would take several minutes. Instead do a normal db migration `sail artisan make:migration` with a single db::statement (or whatever it's called) call.

app/Models/Notification.php
```diff
@@ -10,6 +10,10 @@ class Notification extends Model
 {
     use MassPrunable;
 
+    protected $fillable = [
```
ilestis, line 13: Can be removed if we do the update just using pure sql

ilestis (changes requested):

ilestis (approved):
Magnifico

PR #981, opened 2025-01-09
Objects now have weight

spitfire305:


ilestis (changes requested):
DB tweak, rest looks good!

ilestis (approved):

PR #989, opened 2025-01-16
Fixed map marker bulk edit

spitfire305:


ilestis (changes requested):
Adding an empty option like we have everywhere else for bulk choices seems less convoluted

ilestis (approved):
I love happy endings

PR #990, opened 2025-01-17
Attributes now have to have different names

spitfire305:


app/Http/Requests/UpdateAttributes.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace App\Http\Requests;
+
+use App\Traits\ApiRequest;
+use Illuminate\Foundation\Http\FormRequest;
+
+class UpdateAttributes extends FormRequest
```
ilestis, line 8: Feel like 'SaveAttributes' would be a better name, as this also triggers when creating attributes for the first time?

app/Services/AttributeService.php
```diff
@@ -105,12 +105,11 @@ public function touch(): self
...
     }
 
-    protected function saveAttribute(string $attributeJson): self
+    protected function saveAttribute(array $attributeJson): self
     {
         try {
             /** @var Attribute $attr */
-            $attr = json_decode($attributeJson);
-
+            $attr = json_decode(json_encode($attributeJson));
```
ilestis, line 112: doesn't `json_encode()` cancels `json_encod()`

ilestis (changes requested):
Some tweaks. ALso does this trigger when saving an entity for the first time?

app/Http/Controllers/CrudController.php
```diff
@@ -358,6 +359,12 @@ public function crudStore(Request $request, bool $redirectToCreated = false)
     {
         $this->authorize('create', $this->model);
 
+        $request->validate([
```
ilestis, line 362: This should be in the StoreCreature, StoreCharacter etc rules, since not all entities have attributes. That way it will also be done in all the APIs

app/Http/Controllers/Entities/CreateController.php
```diff
@@ -47,6 +48,12 @@ public function store(StoreCustomEntity $request, Campaign $campaign, EntityType
     {
         $this->authorize('create', [$entityType, $campaign]);
 
+        $request->validate([
```
ilestis, line 51: Needs to be in StoreCustomEntity, not in the crud

app/Http/Controllers/Entities/EditController.php
```diff
@@ -58,6 +59,12 @@ public function index(Campaign $campaign, Entity $entity)
 
     public function save(Request $request, Campaign $campaign, Entity $entity)
     {
+        $request->validate([
```
ilestis, line 62: Needs to be in the request object, not in th econtroller

ilestis (changes requested):
It's still not in the API

ilestis (approved):

PR #993, opened 2025-01-22
Mentions: Show names of private entities

spitfire305:


app/Models/Campaign.php
```diff
@@ -430,6 +430,13 @@ public function getDefaultGalleryVisibilityAttribute(): mixed
         return Arr::get($this->settings, 'gallery_visibility', 'all');
     }
 
+    /**
+     */
+    public function getNamePrivateEntities(): mixed
```
ilestis, line 435: We name all of these helper classes with get...Attribute to make use of Laravel's attribute mapper, that was we can just call name_private_entities. The typing should also be a bool rather than mixed

app/Services/MentionsService.php
```diff
@@ -568,13 +593,28 @@ protected function parseAttributesForEdit(): self
...
+    /**
+     */
+    protected function hiddenEntity(int $id): Entity|null
+    {
+        if (!$this->campaign->getNamePrivateEntities()) {
+            return null;
+        }
+
+        if (!Arr::has($this->entities, (string) $id) && !Arr::has($this->privateEntities, (string) $id)) {
+            $this->privateEntities[$id] = Entity::where(['id' => $id])->withHidden()->first();
```
ilestis, line 612: Can I mention entities from another campaign this way? If so, we need a `['campaign_id' => $this->campaign->id]` in the where

app/Services/MentionsService.php
```diff
@@ -652,6 +692,41 @@ protected function prepareEntities(): void
...
+        }
+
+        // @phpstan-ignore-next-line
+        if (empty($ids)) {
+            return;
+        }
+
+        // Directly get with the mentioned entity types (provided they are valid)
+        // @phpstan-ignore-next-line
+        $entities = Entity::whereIn('id', $ids)->with(['tags:id,name,slug', 'entityType:id,code,is_special'])->withInvisible()->get();
```
ilestis, line 726: Shouldn't be needing tags (the ticket only requires the entity type in the span's data properties)

ilestis (changes requested):

PR #999, opened 2025-01-24
Roles now are a taglike field

spitfire305:


lang/en/campaigns.php
```diff
@@ -197,8 +197,7 @@
         'switch_back_success'   => 'Switched back to your account.',
         'title'                 => 'Members - :name',
         'updates'               => [
-            'added'     => 'Role :role added to :user.',
-            'removed'   => 'Role :role removed from :user.',
+            'roles'     => 'Campaign roles successfully updated for :user',
```
ilestis, line 200: Can just be "Roles" (I'm hoping to move away from the Campaign terminology later this year)

resources/views/components/form/role.blade.php
```diff
@@ -4,12 +4,31 @@
...
+}
+
 $dropdownParent = Arr::get($options, 'dropdownParent');
 $multiple = Arr::get($options, 'multiple');
 ?>
 <select name="{{ $multiple ? 'roles[]' : 'role'}}" id="{{ $multiple ? 'roles' : 'role'}}"
     class="select2 form-role w-100"
     @if (isset($multiple) && $multiple) multiple @endif
-    data-url="{{ route('roles.find', $campaign) }}"
+    data-url="{{ route('roles.find', ['campaign' => $campaign, 'withAdmin' => true]) }}"
```
ilestis, line 28: In the rest of the app we have url params as `with-admin`, `with-abc` so please keep it consistent

app/Http/Controllers/Campaign/MemberController.php
```diff
@@ -65,15 +65,14 @@ public function back(Campaign $campaign)
...
+    public function updateRoles(UpdateUserRoles $request, Campaign $campaign, CampaignUser $campaignUser)
     {
         $this->authorize('update', $campaignUser);
         if (request()->ajax()) {
             return response()->json();
         }
-
         try {
-            $added = $this->service->update($campaignUser, $campaignRole);
+            $added = $this->service->campaign($campaign)->update($campaignUser, $request->get('roles') ?? []);
```
ilestis, line 75: `$request->get('roles', []`) should do the trick

app/Services/Campaign/MemberService.php
```diff
@@ -42,20 +42,23 @@ public function fromRequest(UpdateUserRole $request): self
...
-            ->first();
-
-        // Admin role being switched? Forget the cache
-        if ($campaignRole->isAdmin()) {
-            CampaignCache::campaign($campaignRole->campaign)->clear();
-        }
-
-        // Deleting an existing role
-        if ($role) {
+        foreach ($roleUsers as $role) {
```
ilestis, line 56: I feel like the part where an admin can't remove another admin after the first 15 minutes that they've been added to the admin role (or something like that) is missing here :P

ilestis (changes requested):
Small tweaks and missing admin check

ilestis (approved):
LGTM!

PR #1001, opened 2025-01-28
Icon Suggestion

spitfire305:


resources/views/campaigns/modules/_form.blade.php
```diff
@@ -16,6 +16,13 @@
     <x-forms.field
         field="icon"
         :label="__('campaigns/modules.fields.icon')">
-        <input type="text" name="icon" value="{{ old('icon', $icon) }}" maxlength="40" class="w-full" />
+        <input type="text" name="icon" value="{{ old('icon', $icon) }}" maxlength="40" class="w-full" list="module-icon-list" />
+        <div class="hidden">
+            <datalist id="module-icon-list">
+                @foreach (\App\Facades\MapMarkerCache::iconSuggestion() as $icon)
```
ilestis, line 22: Not needed here

ilestis (changes requested):

app/Models/Bookmark.php
```diff
@@ -121,6 +123,10 @@ class Bookmark extends Model
         'target',
     ];
 
+    protected array $suggestions = [
+        BookmarkCache::class => 'clearSuggestion',
```
ilestis, line 127: Crashes when I hit save

app/Services/Caches/BookmarkCacheService.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     */
+    public function iconSuggestion(): array
+    {
+        $key = $this->iconSuggestionKey();
+        if (Cache::has($key)) {
+            return Cache::get($key);
+        }
+
+        $default = [
```
ilestis, line 23: No need for defaults for bookmarks

app/Services/Caches/EntityAssetCacheService.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     */
+    public function iconSuggestion(): array
+    {
+        $key = $this->iconSuggestionKey();
+        if (Cache::has($key)) {
+            return Cache::get($key);
+        }
+
+        $default = [
```
ilestis, line 23: Just `fa-brands fa-d-and-d-beyond` and `ra ra-aura` are enough suggestions here

app/Services/Caches/EntityAssetCacheService.php
```diff
@@ -0,0 +1,83 @@
...
+            ->pluck('metadata')
+            ->all();
+
+
+        foreach ($settings as $setting) {
+            $data[] = $setting['icon'];
+        }
+
+        foreach ($default as $value) {
+            if (!in_array($value, $data)) {
```
ilestis, line 53: PHPStan correctly says that in a new campaign, $data is never defined

ilestis (changes requested):

ilestis (approved):

PR #1004, opened 2025-01-29
Exporting/Importing custom types is possible now

spitfire305:


app/Services/Campaign/ExportService.php
```diff
@@ -316,33 +364,44 @@ protected function processImage(Image $image): self
 
     protected function process(string $entity, $model): self
     {
+
+        if ($model instanceof Entity) {
+            $modelEntity = $model;
```
ilestis, line 369: $modelEntity is hard to read when in the rest of Kanka we have $entity. Refactor it instead to use $entity, and change the function's $entity variable to $module instead to make it cleaner

app/Services/Campaign/ExportService.php
```diff
@@ -316,33 +364,44 @@ protected function processImage(Image $image): self
...
+        } else {
+            $modelEntity = $model->entity;
+        }
+
         if (!$this->assets) {
```
ilestis, line 374: This if can be removed as we no longer have an asset vs entity export

app/Services/Campaign/ExportService.php
```diff
@@ -316,33 +364,44 @@ protected function processImage(Image $image): self
...
-            $this->archive->add($model->export(), $entity . '/' . Str::slug($model->name) . '.json', );
+            if ($model instanceof Entity) {
+                $this->archive->add(json_encode(['entity' => $model->export(true)]), $entity . '/' . Str::slug($model->name) . '.json', );
```
ilestis, line 376: I think you forgot to remove the (true) param here

app/Services/Campaign/ExportService.php
```diff
@@ -316,33 +364,44 @@ protected function processImage(Image $image): self
...
+            } else {
+                $this->archive->add($model->export(), $entity . '/' . Str::slug($model->name) . '.json', );
```
ilestis, line 378: Also trailing `,` in the add() call which can be removed on both parts

ilestis (changes requested):

app/Services/Campaign/Import/Mappers/CustomEntityMapper.php
```diff
@@ -167,6 +167,24 @@ protected function migrateImage(string $img): Image
         return $image;
     }
 
+    public function tree(): self
+    {
+        dump($this->parents);
```
ilestis, line 172: Mmmhhh

ilestis (changes requested):

app/Services/Campaign/Import/Mappers/CustomEntityMapper.php
ilestis: Also PHPStan complains

ilestis (approved):

PR #1009, opened 2025-02-12
Copying entities now keeps mentions

spitfire305:


app/Models/MiscModel.php
```diff
@@ -126,12 +126,13 @@ public function getLink(string $action = 'show'): string
     /**
      * Create the model's Entity
      */
-    public function createEntity(): Entity
+    public function createEntity(string $newEntry = ''): Entity
```
ilestis, line 129: Entities can have a NULL entry (for example those created in the API), which _might_ cause this call to crash?

app/Services/MentionsService.php
```diff
@@ -624,10 +658,16 @@ protected function parsePostForEdit(string $mention, array $data): string
      */
     protected function entity(int $id): Entity|null
     {
+        Log::info('looking for entity');
```
ilestis, line 661: Logs can be removed

ilestis (changes requested):

ilestis (approved):

ilestis (approved):

PR #1014, opened 2025-03-12
Bookmark confirm modal

spitfire305:


resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -42,4 +43,8 @@
     <x-dialog id="datagrid-filters" :loading="true" full="true">
 
     </x-dialog>
+
+    <x-dialog id="datagrid-bookmark" :loading="true" full="true">
```
ilestis, line 47: No need for another dialog, you can reuse the one that is already on every single page

resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -20,9 +20,10 @@
     @if ($activeFilters > 0)
         @if (empty($bookmark) && isset($entityType))
         @can('create', \App\Models\Bookmark::class)
-            <a href="{{ route('save-filters', [$campaign, $entityType, 'm' => $mode]) }}" class="btn2 btn-sm btn-primary">
-                <x-icon class="fa-solid fa-bookmark" /> {{ __('filters.actions.bookmark') }}
-            </a>
+            <div class="inline-block cursor-pointer btn2 btn-sm btn-primary break-keep" data-toggle="dialog" data-target="datagrid-bookmark" data-url="{{ route('filters.modal_form', [$campaign, $entityType, 'm' => $mode]) }}">
```
ilestis, line 23: Missing aria-label directive for screen readers

resources/views/filters/save_form.blade.php
```diff
@@ -0,0 +1,46 @@
...
+    /** @var \App\Services\SidebarService $sidebar */
+@endphp
+@inject('sidebar', 'App\Services\SidebarService')
+
+<x-form :action="['save-filters', $campaign, $entityType, 'm' => $mode]" method="GET" id="crud-filters-form" class="block">
+<x-dialog.header>
+    {{ __('filters.actions.bookmark') }}
+</x-dialog.header>
+<article>
+    <div class="field flex flex-col gap-3 field-">
```
ilestis, line 11: `field-` seems out of line. gap-3 also seems weird, we don't use gap-3 in any other ui. use <x-grid type="1/1"> instead of the div

ilestis (changes requested):

resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -20,9 +20,9 @@
     @if ($activeFilters > 0)
         @if (empty($bookmark) && isset($entityType))
         @can('create', \App\Models\Bookmark::class)
-            <div class="inline-block cursor-pointer btn2 btn-sm btn-primary break-keep" data-toggle="dialog" data-target="datagrid-bookmark" data-url="{{ route('filters.modal_form', [$campaign, $entityType, 'm' => $mode]) }}">
+            <div class="inline-block cursor-pointer btn2 btn-sm btn-primary break-keep" data-toggle="dialog" data-target="datagrid-filters" data-url="{{ route('filters.modal_form', [$campaign, $entityType, 'm' => $mode]) }}">
```
ilestis, line 23: The div needs a role="button" and the aria-label on it to be accessible to screen readers, otherwise they won't know it's a button.

resources/views/cruds/datagrids/filters/datagrid-filter.blade.php
```diff
@@ -20,9 +20,9 @@
...
                 <x-icon class="fa-solid fa-bookmark" />
-                <span class="hidden sm:inline">{{ __('filters.actions.bookmark') }}</span>
+                <span class="hidden sm:inline"  aria-label="{{ __('filters.actions.bookmark') }}">{{ __('filters.actions.bookmark') }}</span>
```
ilestis, line 25: Not needed here but on the div since a screen reader can read the text inside a span just fine

ilestis (changes requested):

ilestis (approved):

PR #1015, opened 2025-03-13
Organisations: Add members

spitfire305:


app/Http/Controllers/Organisation/MemberController.php
```diff
@@ -84,12 +84,16 @@ public function create(Campaign $campaign, Organisation $organisation)
      */
     public function store(StoreOrganisationMember $request, Campaign $campaign, Organisation $organisation)
     {
+        //here
```
ilestis, line 87: where? 😛

app/Http/Requests/StoreOrganisationMember.php
```diff
@@ -28,7 +28,9 @@ public function rules()
     {
         return $this->clean([
             'organisation_id' => 'required|exists:organisations,id',
-            'character_id' => 'required|exists:characters,id',
+            'characters' => 'required_without:character_id|array|min:1',
```
ilestis, line 31: Create a new request called StoreOrganisationMembers instead of making a complicated one that could potentially break the API

lang/en/organisations.php
```diff
@@ -13,12 +13,14 @@
     ],
     'members'       => [
         'actions'       => [
-            'add'       => 'Add a member',
-            'submit'    => 'Add member',
+            'add'               => 'Add a member',
```
ilestis, line 16: `add` be removed if it's not used anywhere else

lang/en/organisations.php
```diff
@@ -13,12 +13,14 @@
...
+            'submit'            => 'Add member',
```
ilestis, line 17: Same for submit

lang/en/organisations.php
```diff
@@ -13,12 +13,14 @@
...
+            'add_multiple'      => 'Add members',
         ],
         'create'        => [
             'success'   => 'Member added to :name.',
```
ilestis, line 21: Same here, can be removed if no longer used

lang/en/organisations.php
```diff
@@ -13,12 +13,14 @@
...
             'title'     => 'New Member',
```
ilestis, line 22: Same here

app/Http/Controllers/Organisation/MemberController.php
```diff
@@ -84,12 +84,16 @@ public function create(Campaign $campaign, Organisation $organisation)
...
             return response()->json(['success' => true]);
         }
-
-        $relation = OrganisationMember::create($request->all());
+        $count = 0;
+        foreach ($request->get('characters', []) as $character) {
+            $relation = OrganisationMember::create(['character_id' => $character] + $request->except('characters'));
+            $count++;
+        }
         return redirect()->route('entities.show', [$campaign, $organisation->entity])
```
ilestis, line 97: Success should include how many members were added

ilestis (changes requested):

lang/en/organisations.php
spitfire305: Still used

ilestis (approved):

PR #1017, opened 2025-03-18
Upvote Counter

spitfire305:


ilestis (changes requested):
Nice little crash on upvoting an older idea that doesn't have a message_id

ilestis (approved):

PR #1027, opened 2025-03-28
Migrated Observers into models

spitfire305:


ilestis (changes requested):
Didn't ask for this in Kanka

PR #1029, opened 2025-03-31
Redirect to roadmap

spitfire305:


app/Http/Middleware/LoginRedirect.php
```diff
@@ -0,0 +1,26 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+class LoginRedirect
+{
+    public function handle(Request $request, Closure $next): Response
+    {
+        // Store the previous URL in session only if it's not the login route itself and is a valid redirect url
+        if ($request->is('login') || $request->is('register')) {
+            // If to check where the request is coming from and set the variable if its a valid route.
+            if (url()->previous() === route('roadmap')) {
```
ilestis, line 16: Make it a param called `next` in the url, and a whitelist of allowed "next" options, so that it can work from outside links too.

ilestis (changes requested):

ilestis (approved):

PR #1033, opened 2025-04-12
Missing API stuff

spitfire305:


app/Services/AttributeService.php
```diff
@@ -195,7 +195,7 @@ public function updateVisibility(bool $privateAttributes): self
     public function applyEntityTemplates(Entity $entity, int $order = 0): int
     {
         $typeId = $entity->typeId();
-        $templates = AttributeTemplate::has('entity')->where(['entity_type_id' => $typeId])->get();
+        $templates = AttributeTemplate::has('entity')->active()->where(['entity_type_id' => $typeId])->get();
```
ilestis, line 198: Is this supposed to be `enabled()` ?

ilestis (changes requested):
When creating a new location, it's still adding all the attributes from disabled attribute templates.
Also, it needs to show on the attribute template itself if it's disabled

resources/views/entities/components/header.blade.php
```diff
@@ -139,6 +139,12 @@
                     <span class="sr-only">{{ __('characters.hints.is_dead') }}</span>
                 </span>
             @endif
+            @if ($entity->isAttributeTemplate() && !$entity->child->is_enabled)
```
ilestis, line 142: Not what was asked in the ticket

ilestis (changes requested):

ilestis (approved):

PR #1035, opened 2025-04-25
Entities can now be patched from the api

spitfire305:


app/Http/Controllers/Api/v1/EntityApiController.php
```diff
@@ -2,19 +2,22 @@
...
 namespace App\Http\Controllers\Api\v1;
 
+use App\Http\Requests\API\PatchEntity;
 use App\Http\Requests\API\StoreEntities;
 use App\Http\Resources\EntityResource as Resource;
 use App\Models\Campaign;
 use App\Models\Entity;
 use App\Models\EntityType;
 use App\Services\Api\BulkEntityCreatorService;
+use App\Services\Entity\TransformService;
```
ilestis, line 12: I can't figure out where/why this is included

ilestis (changes requested):

app/Http/Controllers/Api/v1/EntityApiController.php
spitfire305: Forgot to remove it

ilestis (approved):

ilestis (changes requested):
Missing "parent_id" for custom modules to make this new endpoint useful

app/Http/Controllers/Api/v1/EntityApiController.php
```diff
@@ -81,8 +81,12 @@ public function patch(PatchEntity $request, Campaign $campaign, Entity $entity)
     {
         $this->authorize('access', $campaign);
         $this->authorize('update', $entity);
-
-        $entity->update($request->only('name', 'type', 'is_private', 'is_template', 'tooltip', 'entry', 'image_uuid', 'header_uuid'));
+        if ($entity->hasChild()) {
```
ilestis, line 84: Please refactor to remove the copy-paste content, otherwise the day we add a field, we 100% will forget to add it in both places

ilestis (changes requested):

app/Http/Controllers/Api/v1/EntityApiController.php
```diff
@@ -76,6 +77,23 @@ public function edit(Request $request, Campaign $campaign, Entity $entity)
         return new Resource($entity);
     }
 
+    public function patch(PatchEntity $request, Campaign $campaign, Entity $entity)
+    {
+        $this->authorize('access', $campaign);
+        $this->authorize('update', $entity);
+        $keys = ['name', 'type', 'is_private', 'is_template', 'tooltip', 'entry', 'image_uuid', 'header_uuid'];
+        if (! $entity->entityType->isSpecial()) {
```
ilestis, line 85: Only special entity types have a parent field so the ! Should be dropped

PR #1067, opened 2025-06-16
Update passport 13

spitfire305:


composer.json
```diff
@@ -28,12 +28,13 @@
...
         "laravel/framework": "^11.0",
         "laravel/pail": "^1.1",
-        "laravel/passport": "^12.0",
+        "laravel/passport": "^13.0",
         "laravel/scout": "^10.5",
         "laravel/socialite": "^5.0",
         "laravel/ui": "^4.6",
         "league/flysystem-aws-s3-v3": "^3.5",
         "league/html-to-markdown": "^5.1",
+        "league/oauth2-server": "^9.2",
```
ilestis, line 37: Why do we need to add this if it's already in laravel/passport's composer.json?

ilestis (changes requested):

ilestis (approved):

PR #1083, opened 2025-06-25
Entity Type logs

spitfire305:


app/Listeners/Campaigns/EntityTypes/LogEntityType.php
```diff
@@ -0,0 +1,46 @@
...
+            $event instanceof EntityTypeDeleted => 'deleted',
+            $event instanceof EntityTypeToggled => 'toggled',
+        };
+
+        if (($event instanceof EntityTypeUpdated && $event->entityType->wasChanged('is_enabled')) || $event instanceof EntityTypeToggled) {
+            $action = $event->entityType->is_enabled ? 'enabled' : 'disabled';
+        }
+
+        $event->user->campaignLog(
+            $event->entityType->campaign_id,
```
ilestis, line 37: Standard entityType objects have no campaign_id, since they are shared by every single campaign

app/Services/Campaign/ModuleEditService.php
```diff
@@ -91,16 +92,7 @@ public function toggle(): bool
...
-            $this->campaign->id,
-            'modules',
-            'toggle',
-            [
-                'id' => $this->entityType->id,
-                'code' => $this->entityType->code,
-                'new' => $this->campaign->setting->{$this->entityType->pluralCode()} ? 'enabled' : 'disabled',
-            ]
-        );
+        EntityTypeToggled::dispatch($this->entityType, auth()->user());
```
ilestis, line 95: Doesn't log anything

app/Services/EntityTypeService.php
```diff
@@ -133,48 +133,13 @@ public function save(): EntityType
         $this->entityType->icon = $this->request->get('icon');
         $this->entityType->save();
 
-        if ($this->entityType->wasRecentlyCreated) {
-            $this->permissions()->bookmark();
```
ilestis, line 137: Why does it no longer create the permissions and the bookmark for the custom module?

ilestis (changes requested):
Bugs

PR #1088, opened 2025-07-01
Infinite scrolling gallery

spitfire305:


ilestis (changes requested):
Doesn't seem to work in folders

PR #1093, opened 2025-07-08
Post layout for children

spitfire305:


resources/views/entities/components/posts/children.blade.php
```diff
@@ -0,0 +1,62 @@
+
+
+@if($entity->hasChild())
```
ilestis, line 3: Remove all of this and make the datagrid init on page load to get rid of any sql query here in the view

ilestis (changes requested):
Remove sql queries from the blade component

PR #1095, opened 2025-07-10
Gallery: replace image

spitfire305:


ilestis (changes requested):
When I try updating a file a second time, it doesn't work

PR #1103, opened 2025-08-11
fixed tests

spitfire305:


app/Http/Resources/CampaignResource.php
```diff
@@ -31,7 +31,7 @@ public function toArray($request): array
         /** @var Campaign $campaign */
         $campaign = $this->resource;
 
-        $url = route('dashboard', $campaign);
```
ilestis, line 34: This is wrong, we use the campaign slug on premium campaigns, not the id

app/Http/Resources/Public/CampaignResource.php
```diff
@@ -17,7 +17,7 @@ public function toArray($request)
             'thumb' => $campaign->image ? $campaign->thumbnail(320, 240) : 'https://th.kanka.io/zzKcBpijSBvm4rPWdzRpI82pTNQ=/320x240/smart/src/app/backgrounds/mountain-background-medium.jpg',
             'name' => $campaign->name,
             'justify' => $campaign->featured_reason,
-            'link' => route('dashboard', $campaign),
```
ilestis, line 20: Same here, the link has the slug, not the id

ilestis (changes requested):
Routing doesn't use campaign->id anymore for premium campaigns with a custom slug

ilestis (approved):

PR #1106, opened 2025-08-13
Attributes bulk api

spitfire305:


ilestis (changes requested):
The alert when trying to save an entity with two attributes having the same name doesn't appear anymore and just deletes the duplicate instead

ilestis (changes requested):
If I make a PATCH request with the following code, it crashes. It shouldn't crash but properly tell the user that the data is invalid.

```json
{
  "attribute": {
    "name": "Attribute A"
  }
}
```

It's also not clear in the docs what the syntax is. Remove the whole need to prefix everything in an `attribute` property.
A valid json should be
```json
{
   [
     "name": "Attribute a",
     "value": "b"
   ],
   [
    "name": "attribute b",
     "value": "c"
   ]
}
```

ilestis (changes requested):
1) Revert all changes to UniqueAttributeNames.
2) Create a new validator just for the api
3) make it properly throw validation errors if the user sends
```json
{
  "attribute": [
    {
      "name": "hello"
    }
  ]
}
```
4) make it generate validation errors if the user sends
```json
{
  "attribute": 
    {
      "name": "hello"
    }
}
```

5) make it generate validation errors if the user sends
```json
{
  "attribute": [
    {
      "name": "hello"
       "type_id": 888888
    }]
}
```
6) make the random type work, or document somewhere why it can't work
```json
{
  "attribute": [
    {
      "name": "hello",
      "value": "1-99",
      "type_id": 5
    }
  ]
}
```

Make it never throw a 500 server error, that's the job of the validator rules to block before it even hits our php code.

ilestis (approved):
LGTM

PR #1253, opened 2026-03-26
Calendar Eras on text

spitfire305:


ilestis (changes requested):
For the third time, we don't need a start/end day or month. **only start year, end year, and name**. Also add support for negative years.

ilestis (changes requested):
It's not showing the relative date in an entry's reminders tab

Image

It's not showing the relative date and era in the calendar's header 

Image

It's not showing the relative date and era in the pinned reminders

Image

It's not showing the relative date and era in the post date?

Image

